PR #232, opened 2014-06-03
cache transform

shinriyo:
Cache transform is better


pharan:
This hasn't been needed at least since 4.5.
https://www.reddit.com/r/Unity3D/comments/2l9tso/transform_is_internally_cached/

Thanks for the contribution!

PR #249, opened 2014-07-09
[Unity] Submesh triangles may not be set when not rendering Meshes.

coeing:
Hi Nathan!
One comment to this return: Is it possible that it should be inside the inner "if" block? We have one issue with a character which doesn't use mesh attachments, but shows some strange behaviour when I turn off the renderMeshes option. When I move the return in the inner if block the issue is gone which makes sense because otherwise the triangles for the sub mesh may never be set if the check "submesh.firstVertex != firstVertex || submesh.triangleCount < triangleCount" fails.

Cheers
Christian


pharan:
The thing this fix seems to do is make (!renderMeshes) not do what it's supposed to do (ie, it will cause it to render meshes anyway.

We will monitor this further with test cases but it's been added to the list of bugs to investigate. This PR will be closed as an invalid fix.

PR #257, opened 2014-07-28
Cocos2d-iphone v.3.1 support

jvasti:
Here are some patches to make spine-cocos2d-iphone work with cocos2d-iphone v. 3.1

PolygonBatch has been removed since Cocos2d-iphone has automatic
batching. CCDrawingPrimitives calls have been changed to CCDrawNode
calls and the rendering of triangles have been ported to CCRenderer.

Xcode projects have been updated with the necessary GLKit dependancy.

The spineboy and goblins examples both seem to be working fine.


spine-cocos2d-iphone/3.0/src/spine/SkeletonRenderer.m
```diff
@@ -210,16 +209,29 @@ - (void) draw {
 			color.r = _skeleton->r * slot->r * r * multiplier;
 			color.g = _skeleton->g * slot->g * g * multiplier;
 			color.b = _skeleton->b * slot->b * b * multiplier;
-			[batch add:texture vertices:worldVertices uvs:uvs verticesCount:verticesCount
-				triangles:triangles trianglesCount:trianglesCount color:&color];
+            self.texture = texture;
+            CGSize size = texture.contentSize;
+            GLKVector2 center = GLKVector2Make(size.width/2.0, size.height/2.0);
+            GLKVector2 extents = GLKVector2Make(size.width/2.0, size.height/2.0);
+            if(CCRenderCheckVisbility(transform, center, extents)){
```
NathanSweet: Sorry it's been quite some time, but what is this `CCRenderCheckVisbility` call doing? To determine if the vertices are on screen, we'd need to find the min and max vertices for x and y. This isn't related to the size of the texture, which is likely a texture atlas.

PR #267, opened 2014-08-05
Listener class improvements

NimaiMalle:
Changed the add function for Listeners to not add the same listener function more than once.
Changed the remove function for Listeners to not delete the last listener in the list if the requested listener isn't present in the list.
Added a "has" function to test whether or not a listener function already exists in the list of listeners.


badlogic:
This has been fixed, thanks for your contribtution!

PR #271, opened 2014-08-21
[lua] Bugfixes and (some) improvements to spine-lua

brocoli-tapps:
The most important commits are the folowing:
- cb9c590 : this was a method call with "." instead of ":", that caused a crash.
- 6bd6d45 : the tracks array had holes in it. The iteration loop was correct -- that is, it used a field .trackCount and tested for the existence of each tracks[i] -- but something was wrong with the updating of the .trackCount variable. It seemed more simple to just iterate through the tracks table with pairs() instead, and scrap the .trackCount field entirely, and with it, the need to update it by hand. This commit has a lenghty diff because of the change in identation this caused.
- 80a0d31 : the frames table uses the 0-th field, this is not part of a table's array in Lua, and because of this the frameCount variable was always counting one frame less than it should. Since the rest of the function assumes that frames[0] always exists, it was safe to simply increment frameCount by 1.

the other commits are simple maintenance: removing unused code, fixing file permissions, and replacing a hard-coded linear search with an already existing binarySearch function.


badlogic:
I've now rewired all kinds of things Lua and localized as much as possible. See the impending 3.5 release.

PR #285, opened 2014-09-11
findAnimation method for cocos2d-x

newnon:
useful to determinate duration of animation


badlogic:
Thanks, added manually.

PR #298, opened 2014-09-25
Spine c-runtime struct definition & sfml memory access after free

pinguin999:
Spine c-runtime struct definition
sfml memory access after free


NathanSweet:
Thanks!

PR #311, opened 2014-10-09
Clamp lastTime to endTime

fillrate:
Currently, for a non-looping animation, lastTime is not clamped to the length of the animation but time is.  This means then when applying an animation, time could be less than lastTime and this in turn can lead
to all events in the animation being fired every frame, when the animation has reached the end.

This commit clamps the lastTime in a similar manner to the time, when applying animations.

This is only fixed in the spine-c runtime, as I'm not in a position to test the others.  I did notice that the same issue will occur in other runtimes however.

### DIFF
--- a/spine-c/src/spine/AnimationState.c
+++ b/spine-c/src/spine/AnimationState.c
@@ -119,7 +119,7 @@ void spAnimationState_apply (spAnimationState* self, spSkeleton* skeleton) {
 	int i, ii;
 	int eventsCount;
 	int entryChanged;
-	float time;
+	float time, lastTime;
 	spTrackEntry* previous;
 	for (i = 0; i < self->tracksCount; ++i) {
 		spTrackEntry* current = self->tracks[i];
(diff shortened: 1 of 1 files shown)


NathanSweet:
I'm having trouble following. When would `time` be erroneously less then `lastTime`?

fillrate:
It's a while since I looked at this, so forgive me if I'm a bit wooley.

I believe the bug surfaced after playing a non-looping animation containing events, followed by another animation with a delay.  The delay was longer than the length of the first animation, so both `current->lastTime` and `current->time` became greater than `current->endTime` for the tracks.

In `spAnimationState_apply` it uses the temporary value `time`, set to `minimum(current->time, current->endTime)` for non-looped animations.  In the case described above, where both `current->lastTime` and`current->time` are greater than `current->endTime`, `time` will always be less than `current->lastTime`.

Those times are passed to the `spAnimation_apply` and `spAnimation_mix` functions and in the case of the event timeline (in `_spEventTimeline_apply`), this effectively becomes;

* fire all events from `current->lastTime` to `INT_MAX` - this does nothing as `current->lastTime` is greater than the last frames time and it does an early-out.
* fire all events from -1 to `current->endTime` - this fires all the events in the timeline

My fix simply makes those calls in `spAnimationState_apply` use `minimum(current->lastTime, current->endTime)` instead of `current->lastTime` for non-looped animations.  I did contemplate setting `current->lastTime` to the clamped `time` value at the end of the loop instead of `current->time`, but `spAnimationState_update` relies on `lastTime` going greater than `endTime`, when there are pending next animations.

badlogic:
A lot has changed since this PR was opened. Sorry for us talking so long to get to it. Please give the latest spine-c runtime a try. If your issue is not fixed, we can re-open. Thanks for your contribution.

PR #325, opened 2014-11-18
[C + AS3] fix bug may be happened during mix animations

zhongfq:
during mix period and in the same slot, if pervious timeline has attachment change after the current timeline attachment change, the slot attachment may be not correct after mix period.


badlogic:
This has been fixed.

PR #326, opened 2014-11-19
Optional array offset for coputeWorldVertices in Attachements

asRIA:
Motivation to create this change is optimization and keep parity with spine-c. In case of C we can just pass pointer to array. That means we cant update any part of biggest memory block. 

This feature is one of core feature to improve performance, and create one mesh spine object.

This change doesn't break API


badlogic:
This has been fixed in the latest spine-as updates.

PR #329, opened 2014-11-23
Why not link to spine.js?

makc:
Instead of place-it-here.txt? I know, you could say, it will not work on some systems (windows), but then txt file does not make it work either. Whereas on linux and mac this would work out of the box.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,5 +90,3 @@ spine-as3/spine-as3-example/bin-release
 spine-starling/spine-starling/bin
 spine-starling/spine-starling-example/bin-debug
 spine-starling/spine-starling-example/bin-release
-
-spine-turbulenz/spine-js/spine.js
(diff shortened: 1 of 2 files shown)


NathanSweet:
What does Git do when pulling the symlink on Windows? It would be confusing to have a `spine.js` file that isn't a symlink and also isn't the `spine.js` file from spine-js. From some Googling, it seems symlinks in Git can cause trouble.

There are a number of places in this repo where one runtime requires the contents of another to be copied into it. A place-it-here file isn't great, but I'm not sure committing symlinks is a better solution.

makc:
It is not better solution on Windows, you are right about it. But it is better on Mac OS X and all those Linux-es that geek devs love so much.

NathanSweet:
I'm ok with doing things to make things better for those guys, but in this case I think it has too much of a negative impact on Windows users. They are people too! ;)

PR #334, opened 2014-12-01
[JAVASCRIPT] Atlas attachment loader fix.

darionco:
Atlas attachment loader was using the name as the path to the images in the atlas instead of the path argument passed to it. It would throw an exception using the path argument to report the error, making it extremely confusing. It now properly uses the path argument to look for the image.

This bug made it a requirement to have images and attachments that held them have the same name, not very visible bug in spine because the editor names things properly most of the time but with a more strict naming convention this bug is very evident.

I believe that this bug is present in other runtimes, further investigation is needed.


NathanSweet:
Ouch, good find. I checked the other runtimes and they are ok. Cheers!

PR #337, opened 2014-12-07
multiple atlas support, custom start and end time for play/loop animations

gonchar:
- multiple atlases support for spine animations with a lot of graphics
- now you can specify custom start and end time for animation play/loop
- small error fixes


badlogic:
This has all been superseded by the latest spine-as runtime updates. Thanks for your contribution!

PR #338, opened 2014-12-16
Fix issue with HitTest

Fraggle:
Hi,

I dont think you need to apply the scales here because it will be done automatically by the framework (the localPoint is already transformed in the locales coordinates)

Have a nice day!

Seb


NathanSweet:
Ok, thanks! I am guessing it wasn't working for your scaled skeleton and this fixed it?

Fraggle:
Yes indeed :)

PR #342, opened 2015-01-16
Fix buffer overflow when JSON Skeleton produces Error

asRIA:
Missing space for `\0`
Problem has been highlighted during loading invalid json.


NathanSweet:
Thanks! This must be the easiest and most common mistake to make in the world. :smile:

PR #344, opened 2015-01-19
[C#] Optional fix for firing events when mixing

pharan:
Not sure if the same "fix" needs to be applied to other runtimes, or if this was intended behavior at all. But someone in the forums was looking for this fix.


pharan:
This allows previous animations to fire its events during mixing/crossfading.

To maintain consistency across runtimes, this fix code is commented out.
To use it, remove the old line and uncomment the fix line.

PR #345, opened 2015-01-21
[Cocos2dx 2] Schedule update calls when entering the scene.

j-jorge:
If a node is removed from the scene, then Cocos2d-x will unschedule its
selectors. In order to have the animation still playing if the node is
inserted again in the scene, the call to scheduleUpdate() in SkeletonRenderer
must then be done in SkeletonRenderer::onEnter() instead of at initialization
time.

This commit applies this change, making the following code to work:

    void testAnimatedWhenReinserted( SkeletonRenderer& skeleton )
    {
        skeleton.retain();
        cocos2d::CCNode* const parent{ skeleton.getParent() };
        skeleton.removeFromParent();
        parent->addChild( &skeleton );
        skeleton.release();
    }


badlogic:
We no longer maintain the spine-cocos2dx runtime for Cocos2D-X v2 I'm afraid. We have a branch with the latest version of that runtime here: https://github.com/esotericsoftware/spine-runtimes/tree/cocos2dx-v2

Thanks for your contribution.

PR #366, opened 2015-02-19
[lua] Fixed "addAnimation" function not triggering "onStart" callback when queue was empty

brunopezzolo:
When calling AnimationState's "addAnimation" function and the AnimationsStates's queue was empty, the animation's "onStart" trigger would fail to be called, even though the animation would still run.


pharan:
[runtimes] AnimationState Start event not fired when using AddAnimation on empty track.
https://github.com/EsotericSoftware/spine-runtimes/pull/366

NathanSweet:
Should be fixed in the runtimes for 3.5.

PR #368, opened 2015-02-20
[C#] Updated AttachmentTimeline to check whether expected attachment is set.

Cookingsource:
Until now it was discarding the set of the attachment if lastTime is after keyTime. I'm having issues with this behavior when the skeleton is set to pose manually. After that, the attachment is the one used in setup intead of the expected one until the next keyframe.

For more information on the scenario. I'm using this to revert to pose only specific tracks. Doing the following.
1 -  Clear the Track To Revert
2 -  Revert Skeleton to pose
3 -  Update(0) this way the rest of the tracks apply to the skeleton again

I've made several characterization tests of this timeline before changing it, I can pass them to you if you want.

### DIFF
--- a/spine-csharp/src/Animation.cs
+++ b/spine-csharp/src/Animation.cs
@@ -447,12 +447,20 @@ namespace Spine {
 				lastTime = -1;
 
 			int frameIndex = (time >= frames[frames.Length - 1] ? frames.Length : Animation.binarySearch(frames, time)) - 1;
-			if (frames[frameIndex] < lastTime) return;
-
-			String attachmentName = attachmentNames[frameIndex];
-			skeleton.slots[slotIndex].Attachment =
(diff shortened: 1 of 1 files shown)


pharan:
This is an ongoing discussion on how to implement efficient overriding behavior for attachment timelines. Overriding (in multi-track cases) is desired behavior.

The current implementation does not do this, and causes `AttachmentTimeline` to behave like an event timeline, only applying its changes during the frame when time passes over the key but not any time after.

Getting the overriding behavior is as easy as commenting out this check:
```csharp
if (frames[frameIndex] < lastTime) return;
```

There were recent changes in `skin.GetAttachment` (removal of string comparisons) which was what the current implementation was trying to avoid. So the overidding behavior may just be officially restored by removing that line.

PR #369, opened 2015-02-21
add C++ constructors to all structs (VS Compiler C++ compatibility)

jrenner:
for Visual Studio compiler compatibility where projects are using C++

----------------

spine-c now compiles successfully in VS 2013!

I don't know what I'm doing with regards to what values to use for initialization.  most things are ints, floats, chars or pointers so I set them to 0 and hope that it will all work out somehow. I think they will all be overwritten by your custom OO create methods but I'm not sure how that all works. Maybe take a cursory glance at that? Also the super() inits, did I do those right?

### DIFF
--- a/spine-c/include/spine/Animation.h
+++ b/spine-c/include/spine/Animation.h
@@ -47,6 +47,15 @@ typedef struct spAnimation {
 
 	int timelinesCount;
 	spTimeline** timelines;
+
+#ifdef __cplusplus
+	spAnimation() :
+		name(0),
+		duration(0),
+		timelinesCount(0),
+		timelines(0) {
+	}
+#endif
 } spAnimation;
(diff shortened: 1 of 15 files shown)


spine-c/include/spine/Animation.h
```diff
@@ -91,8 +100,14 @@ typedef enum {
 
 struct spTimeline {
 	const spTimelineType type;
-
 	const void* const vtable;
+
+#ifdef __cplusplus
+	spTimeline() :
+		type(SP_TIMELINE_SCALE),
```
NathanSweet: Can be zero. All const struct fields are set by code in the `create` or `init` OO functions.

NathanSweet:
Looks good!

PR #370, opened 2015-02-21
Fixed cocos2d-x Primitives deprecation by moving to DrawNode

Mazyod:
Hey,

I recently updated to the latest cocos2d-x version, and got a bunch of annoying warnings from `DrawPrimitives` used in the spine library for debugging nodes.

I simply refactored that code to use `DrawNode`, then moved it to the draw method, so we can pass the renderer to the `DrawNode` directly.

According to the docs, `DrawNode` should be available since `2.1`, so backward compatibility should be OK.

Thanks @NathanSweet


badlogic:
Thanks, I ported it to the latest spine-cocos2dx runtime!

PR #374, opened 2015-02-24
Added Sorting Layers GUI to SkeletonRenderer and SkeletonAnimation

pharan:
Added Sorting Layers interface to Spine-Unity SkeletonRenderer and SkeletonAnimation editor inspectors.
Also added foldout/collapsible for the advanced options like Calculate Normals, Front Facing, etc...

### DIFF
--- a/spine-unity/Assets/spine-unity/Editor/SkeletonAnimationInspector.cs
+++ b/spine-unity/Assets/spine-unity/Editor/SkeletonAnimationInspector.cs
@@ -66,6 +66,7 @@ public class SkeletonAnimationInspector : SkeletonRendererInspector {
 			}
 		}
 
+		EditorGUILayout.Space();
 
 		//TODO:  Refactor this to use GenericMenu and callbacks to avoid interfering with control by other behaviours.
 		// Animation name.
@@ -80,10 +81,7 @@ public class SkeletonAnimationInspector : SkeletonRendererInspector {
 					animationIndex = i + 1;
 			}
 		
-			EditorGUILayout.BeginHorizontal();
-			EditorGUILayout.LabelField("Animation", GUILayout.Width(EditorGUIUtility.labelWidth));
-			animationIndex = EditorGUILayout.Popup(animationIndex, animations);
-			EditorGUILayout.EndHorizontal();
+			animationIndex = EditorGUILayout.Popup("Animation", animationIndex, animations);
 
 			String selectedAnimationName = animationIndex == 0 ? null : animations[animationIndex];
 			if (component.AnimationName != selectedAnimationName) {
--- a/spine-unity/Assets/spine-unity/Editor/SkeletonRendererInspector.cs
+++ b/spine-unity/Assets/spine-unity/Editor/SkeletonRendererInspector.cs
@@ -37,8 +37,9 @@ public class SkeletonRendererInspector : Editor {
 	protected SerializedProperty skeletonDataAsset, initialSkinName, normals, tangents, meshes, immutableTriangles, submeshSeparators, front;
 	protected bool advancedFoldout;
 
-	private static PropertyInfo SortingLayerNamesProperty;
-	private static MethodInfo GetSortingLayerUserID;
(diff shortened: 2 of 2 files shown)


NathanSweet:
@Fenrisul Looks good?

LostPolygon:
This won't work on Unity 5, since Unity 5 has no concept of "user layer ID", instead it uses "unique layer ID". It's probably best to use UnityEditor.EditorGUILayout.SortingLayerField (it's internal, but works fine via reflection).

pharan:
@LostPolygon Thanks for the tip. That works fine in Unity 4.x too, I assume?

LostPolygon:
Yeah, except, if I recall it right. the name of SerializedProperty for the layerID is different. Just grab a decompiler and examine UnityEditor.SpriteRendererInspector.OnInspectorGUI method to see exactly how Unity editor uses it.

pharan:
@LostPolygon That's the problem I'm running into right now. It's pretty annoying. Might not be able to get to do this today.

pharan:
@LostPolygon So I got the property to draw using a fake sortinglayerid property field. Code's much cleaner.

My problem right now is that I actually need to access the SerializedObject of the MeshRenderer from SkeletonRenderer's Editor/Inspector code. How do you do that?

pharan:
That should do it. Thanks again, @LostPolygon

pharan:
Don't merge this yet. It's broken in 5. No idea how to fix it.

```
typeof(EditorGUILayout).GetMethod("SortingLayerField", BindingFlags.Static | BindingFlags.NonPublic);
```
returns null and throws an exception:

```
AmbiguousMatchException: Ambiguous matching in method resolution
System.Reflection.Binder.FindMostDerivedMatch (System.Reflection.MethodBase[] match) (at /Users/builduser/buildslave/mono-runtime-and-classlibs/build/mcs/class/corlib/System.Reflection/Binder.cs:106)
System.MonoType.GetMethodImpl (System.String name, BindingFlags bindingAttr, System.Reflection.Binder binder, CallingConventions callConvention, System.Type[] types, System.Reflection.ParameterModifier[] modifiers) (at /Users/builduser/buildslave/mono-runtime-and-classlibs/build/mcs/class/corlib/System/MonoType.cs:245)
System.Type.GetMethod (System.String name, BindingFlags bindingAttr) (at /Users/builduser/buildslave/mono-runtime-and-classlibs/build/mcs/class/corlib/System/Type.cs:787)
SkeletonRendererInspector.OnEnable () (at Assets/spine-unity/Editor/SkeletonRendererInspector.cs:76)
SkeletonAnimationInspector.OnEnable () (at Assets/spine-unity/Editor/SkeletonAnimationInspector.cs:41)
```

I tried walking through EditorGUILayout's methods. SortingLayerField isn't in its list anymore.
Checked with a decompiler and the method is still there. I just can't get to it for some reason.

@LostPolygon 
Any guesses?

LostPolygon:
@pharan 
Yes, there are multiple overloads for SortingLayerField method, and you only need one of them. Overloads are determined by the types of parameters. So it should be something like
```
typeof(EditorGUILayout).GetMethod("SortingLayerField", BindingFlags.Static | BindingFlags.NonPublic, null, new Type[] { typeof(GUIContent), typeof(SerializedProperty), typeof(GUIStyle) }, null);
```

pharan:
@LostPolygon
Oh damn. Yeah, I just realized I was inspecting the wrong dll. I only installed the 64-bit Unity and the 32-bit one was still at 4.6.

Thanks loads! I'll be making this change today.

pharan:
@LostPolygon
Oh damn. Yeah, I just realized I was inspecting the wrong dll. I only installed the 64-bit Unity and the 32-bit one was still at 4.6.

Thanks loads! I'll be making this change today.

pharan:
There. Confirmed working in Unity 5.

NathanSweet:
Ready to merge then?

pharan:
I say it's ready.

PR #375, opened 2015-02-26
A bunch of optimizations

LostPolygon:
Replaced most List<T> with ExposedList<T> that exposes the inner array in order to avoid the overhead of using the indexer. Manually inlined some stuff in SkeletonRenderer, and did some other microoptimizations.
For me, all this gave a 15-25% performance boost both on desktop and Android, in a stress-test of 70 characters that use bounding box attachments and meshes with FFD.


pharan:
Those are impressive numbers.

1) Do those numbers also include the effect of reduced method calls?
2) Does passing through struct initializers really cost more than assigning the struct members one by one?
3) Who wrote ExposedList<T>? You?
@NathanSweet has some licensing stuff you might want to take care of either way.
4) Definitely yes on the `is + as` to `as + null check`. I've been wanting to make that change for a while already, but I wasn't sure how much of a difference it made.

LostPolygon:
1) Yes, these numbers are for all the combined changes done in these commits.
2) Oh yes, yes it does. It really _shouldn't_, but Mono JIT is really bad at inlining stuff. This results in a method invocation and a struct copy, and that becomes really noticeable in large loops.
3) I've been using this class for years, but now when you mention it... I think I originally grabbed it from .NET [reference source](http://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,aa7e01fcb80a917e) and then removed like 90% of it and adapted to my needs. This is likely a problem since the [license](http://referencesource.microsoft.com/license.html) forbids publicly distributing modified version. Oops. I think I shall use List<T> from Mono project instead as a base?
4) I don't think this actually makes a big enough change as well, but I did it since I was there anyway.

NathanSweet:
Interesting!

Not using the standard List means the lists are not compatible for sorting and other usage. This is pretty unfortunate.

A 15-25% increase on Android would typically equate to a very small increase on desktop. You are really seeing that big of a jump on desktop also? Can someone else examine performance before and after this PR on desktop and Android, just to double check?

Would you be so kind as to sign our CLA?
http://esotericsoftware.com/files/cla.txt
The list class needs to be unencumbered by other licenses, or use a license compatible with the Spine runtimes license. It's not rocket surgery though. We could write something based on my [Array class](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/Array.java).

LostPolygon:
I'll sign and send the CLA in a few hours.

I've replaced ExposedList with a MIT-licensed version from Mono project, is that ok license-wise? Not really sure if it is compatible with Spine runtime license.
This version is now feature-identical to a regular list (except the indexer, obviously), so it should be easier to use. Actually, the only place I've found where it is impossible to replace List by ExposedList are public properties serialized by Unity, since Unity serializer has no support for generics, with a special exception of regular List<T>.

I've also noticed that SkeletonRenderer uses Mesh.RecalculateBounds() to calculate the mesh bounds. But since we are generating the mesh manually, it'd make sense to calculate bounds manually as well - that's easy enough. The results are quite surprising. Performance gain from doing that is almost unnoticeable on Android (not actually sure why, though), but on desktop, this gave an additional ~30% boost out of a sudden, and I've checked that on two very different PCs. So, in my case all my optimizations in total gave me around 40-50% more performance on desktop, which is pretty neat :) It'd probably be not that much of a gain if no mesh attachments are used, though.

Here's a very simple desktop stress-test in two version, one compiled with all my optimizations and second is without:
https://www.dropbox.com/s/vth378cnwyp55bf/spine_optimizations_test.zip?dl=0
Note that there is someting like a hundred identical characters actually, but all of them have MeshRenderer disabled (except one, for debugging purposes) just to make sure GPU won't interfere with FPS. But yeah, it'd be nice if someone else examine this for different cases, as I've only tested all this with a single character animation and on quite powerful devices.

NathanSweet:
Avoiding `RecalculateBounds` sounds good!

Totally weird you're seeing performance gains on desktop and not Android. Up is down, red is green...

LostPolygon:
CLA signed and sent.

LostPolygon:
Optimized some more stuff. The most important part is that SkeletonRenderer.AddSubmesh and Mesh.SetTriangles are only called when any attachment or a submesh has changed. Only vertices and vertex attributes are updated otherwise. In my tests, this almost doubled the FPS on desktop compared to the yesterday state of things in this PR. To sum it all up, this whole PR now almost triples the FPS for me on desktop. Performance gain is much smaller on Android, for some reason - whole PR gives me around 33% more FPS.
Well, I think I'm actually finished here, I can't see anything else really obvious to optimize that can actually bring a noticeable performance gain :)

spine-csharp/src/Skin.cs
```diff
@@ -75,27 +75,39 @@ public class Skin {
...
 		// Avoids boxing in the dictionary.
-		private class AttachmentComparer : IEqualityComparer<KeyValuePair<int, String>> {
-			internal static readonly AttachmentComparer Instance = new AttachmentComparer();
+		private class AttachmentKeyTupleComparer : IEqualityComparer<AttachmentKeyTuple> {
+			internal static readonly AttachmentKeyTupleComparer Instance = new AttachmentKeyTupleComparer();
 
-			bool IEqualityComparer<KeyValuePair<int, string>>.Equals (KeyValuePair<int, string> o1, KeyValuePair<int, string> o2) {
-				return o1.Key == o2.Key && o1.Value == o2.Value;
+			bool IEqualityComparer<AttachmentKeyTuple>.Equals (AttachmentKeyTuple o1, AttachmentKeyTuple o2) {
+			    return o1.SlotIndex == o2.SlotIndex && o1.NameHashCode == o2.NameHashCode;
```
NathanSweet: It's a nice short circuit but a hashcode check is not the same as equals. We still need an equality check if the hashcodes are equal.

spine-unity/Assets/spine-unity/SkeletonAnimator.cs
```diff
@@ -202,4 +204,14 @@ public enum MixMode { AlwaysMix, MixNext, SpineStyle }
 			_UpdateComplete(this);
 		}
 	}
+
+	private int GetAnimationClipNameHashCode(AnimationClip clip) {
+		int clipNameHashCode;
+		if (!clipNameHashCodeTable.TryGetValue(clip, out clipNameHashCode)) {
```
NathanSweet: Is a method call and hash map look up really faster than computing a string hashcode? I would expect string internally caches the hash code (Java does, FWIW).

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -254,80 +299,163 @@ public class SkeletonRenderer : MonoBehaviour {
...
+				if (tempVertices[RegionAttachment.Y4] < meshBoundsMin.y)
+					meshBoundsMin.y = tempVertices[RegionAttachment.Y4];
+				else if (tempVertices[RegionAttachment.Y4] > meshBoundsMax.y)
+					meshBoundsMax.y = tempVertices[RegionAttachment.Y4];
+
+				// Calculate min/max Z
+				if (z < meshBoundsMin.z)
+					meshBoundsMin.z = z;
+				else if (z > meshBoundsMax.z)
+					meshBoundsMax.z = z;
```
NathanSweet: We could move `meshBoundsMin/Max.z` outside the loop and use the number of visible attachments to compute it. `z` is `i * zSpacing` but it could change to `i * visibleAttachmentCountSoFar`.

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -254,80 +299,163 @@ public class SkeletonRenderer : MonoBehaviour {
...
+						else if (tempVertices[ii] > meshBoundsMax.x)
+							meshBoundsMax.x = tempVertices[ii];
+						if (tempVertices[ii + 1]< meshBoundsMin.y)
+							meshBoundsMin.y = tempVertices[ii + 1];
+						else if (tempVertices[ii + 1] > meshBoundsMax.y)
+							meshBoundsMax.y = tempVertices[ii + 1];
+						if (z < meshBoundsMin.z)
+							meshBoundsMin.z = z;
+						else if (z > meshBoundsMax.z)
+							meshBoundsMax.z = z;
```
NathanSweet: Same here about moving `meshBoundsMin/Max.z` outside the loop.

spine-unity/Assets/spine-unity/SkeletonAnimator.cs
LostPolygon: .NET sure does not cache the string hash code. Just looked at Mono source to be extra sure, and no, there is no caching. 
The whole hashmap thing is to avoid allocating new strings when accessing `AnimationClip.name`, so it _might_ be a tiny bit slower, but reduces the amount of generated garbage.

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -254,80 +299,163 @@ public class SkeletonRenderer : MonoBehaviour {
...
+							else if (tempVertices[ii] > meshBoundsMax.x)
+								meshBoundsMax.x = tempVertices[ii];
+							if (tempVertices[ii + 1]< meshBoundsMin.y)
+								meshBoundsMin.y = tempVertices[ii + 1];
+							else if (tempVertices[ii + 1] > meshBoundsMax.y)
+								meshBoundsMax.y = tempVertices[ii + 1];
+							if (z < meshBoundsMin.z)
+								meshBoundsMin.z = z;
+							else if (z > meshBoundsMax.z)
+								meshBoundsMax.z = z;
```
NathanSweet: Another `z`.

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -366,9 +499,85 @@ public class SkeletonRenderer : MonoBehaviour {
...
+			lastState.addSubmeshArgumentsMesh2;
+
+		// Check attachments
+		int attachmentCount = attachmentsTriangleCountTemp.Count;
+		if (attachmentsTriangleCountCurrentMesh.Count != attachmentCount) {
+			mustUpdateMeshStructure = true;
+		} else {
+			for (int i = 0; i < attachmentCount; i++) {
+				if (attachmentsTriangleCountCurrentMesh.Items[i] != attachmentsTriangleCountTemp.Items[i]) {
+					mustUpdateMeshStructure = true;
```
NathanSweet: Why not just return true here? :)

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -473,24 +689,59 @@ public class SkeletonRenderer : MonoBehaviour {
...
+			public AddSubmeshArguments(Material material, int startSlot, int endSlot, int triangleCount, int firstVertex, bool lastSubmesh) {
+				this.material = material;
+				this.startSlot = startSlot;
+				this.endSlot = endSlot;
+				this.triangleCount = triangleCount;
+				this.firstVertex = firstVertex;
+				this.lastSubmesh = lastSubmesh;
+			}
+
+			public bool Equals(AddSubmeshArguments other) {
```
NathanSweet: Would this be better passed by ref?

spine-csharp/src/ExposedList.cs
```diff
@@ -0,0 +1,583 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Spine {
+	[Serializable]
+	[DebuggerDisplay("Count={Count}")]
+	public class ExposedList<T> : IEnumerable<T> {
```
NathanSweet: This class name is in the code a *lot*. Any ideas for a better name? I like `Array` and use that in libgdx/Java-land. The runtime libs (both C# and Java) have a class named `Array` but it rarely appears in user code.

spine-unity/Assets/spine-unity/SkeletonAnimator.cs
NathanSweet: `AnimationClip.name` allocates? Yeesh...

NathanSweet:
That is quite a big performance jump, even on Android! Great stuff, thanks for taking the time to go through it all! I made a few mostly useless, nitpicking comments. ;) I don't have time at the moment to take it for a test drive, but assuming it works as advertised, it looks good to me. @Fenrisul @pharan, you guys want to weigh in?

spine-csharp/src/ExposedList.cs
LostPolygon: I like `Array` as a name in general, but I actually use it quite often in user code, and it's also confusing because it bears no resemblance with regular `List`, so `Array` is a personal 'no'. Ideas: `ExList`, `RawList`. It's actually called `NanoList` in my other projects, but that's a stupid name :)

Fenrisul: NGUI uses "BetterList" heh

LostPolygon:
No problem. I've fixed those minor issues you've pointed out, and checked again if everything works. It seems like it does, but to be honest, I _do_ have a gut feeling I've accidentally broke _something_ with all these changes all over place :)

spine-csharp/src/ExposedList.cs
NathanSweet: I guess with C#'s odd usage of start caps for methods and properties `Array` might show up more often in user code.

`BetterList` is worse. :p

If we want to keep the word "List" in the name there aren't so many options. `ArrayList` makes sense but seems to be used by the runtime lib. `ListArray`, `ExList`, `RawList`, `BareList`, `List2`, `List_`... maybe not so great. I don't know, I guess I'm fine with `ExposedList`.

NathanSweet:
I think it looks good. It was pretty complicated before and certainly didn't get any simpler, but we can work out any bugs as they are reported. Let's give any potential naysayers a day or two, then merge. Cheers!

LostPolygon:
Sooo... No naysayers in 4 days? Merge time!

Fenrisul:
Too busy at gdc

NathanSweet:
Yea yea, rub GDC in our faces! Let's give @Fenrisul a bit longer. Last day of GDC is the 6th.

LostPolygon:
GDC... I wish I could use that excuse without making that a lie, lol.

Fenrisul:
I gave SpriteBuilder's booth a raft of **** for not being able to do all
the things Spine could do after they claimed it could.

NathanSweet:
Haha! Get 'em! ;)

@Fenrisul maybe you can stop by booth #1024 South and ask the UE guys about Spine support? I sent them an email about their Unreal Dev Grant. It would be great to support UE!

Fenrisul:
Giving this a good lookover today.

Fenrisul:
Broken issue with SubmeshRenderer.
Optimization to disable updating mesh when MeshRenderer is disabled prevents SkeletonSubmeshRenderer from updating as well.

Fenrisul:
Also looks like FrontFacing is broken now too.

spine-unity/Assets/spine-unity/SkeletonRenderer.cs
```diff
@@ -413,12 +605,17 @@ public class SkeletonRenderer : MonoBehaviour {
...
+			Slot slot = drawOrder.Items[i];
 			Attachment attachment = slot.attachment;
 			Bone bone = slot.bone;
-			bool flip = frontFacing && ((bone.WorldFlipX != bone.WorldFlipY) != (Mathf.Sign(bone.WorldScaleX) != Mathf.Sign(bone.WorldScaleY)));
+
+			bool worldScaleXIsPositive = bone.worldScaleX >= 0f;
+			bool worldScaleYIsPositive = bone.worldScaleY >= 0f;
+			bool worldScaleIsSameSigns = (worldScaleXIsPositive && worldScaleYIsPositive) || 
+										 (!worldScaleXIsPositive && !worldScaleYIsPositive);
+			bool flip = frontFacing && ((bone.worldFlipX != bone.worldFlipY) != worldScaleIsSameSigns);
```
Fenrisul: should be == worldScaleIsSameSigns

LostPolygon:
@Fenrisul 
I'm a bit busy at the moment, I'll examine this in a few days

NathanSweet:
Would love to merge this! :) It doesn't merge automatically, but I guess the biggest problem is the front facing stuff broke.

LostPolygon:
Sorry guys, totally forgot about this :(
Giving a look at the front facing and SubmeshRenderer stuff. I personally neveer used them, so I have to figure out what they were for in the first place...

LostPolygon:
Did my best on fixing the front facing and SubmeshRenderer stuff. @Fenrisul, could you give it a look?

Fenrisul:
Yea - I'll give it a look tomorrow.  Was at silicon valley VR conf all day today heh.

I'll have to go over it w/ Nate directly for a bit afterward anyway, so it might be a week or two before we merge it in, but yea. Thanks for cleaning it up some :)  I wouldn't mind an extra 40% performance heh.

LostPolygon:
@Fenrisul Not to nag you, but did you had a chance of looking into it?

Fenrisul:
A bit... but Nate's gone until June 10th...  going to have to wait until
then either way.

LostPolygon:
ping

NathanSweet:
I'd like to get this merged, just @Fenrisul and I haven't had a change to sit down and go over it. Currently it can't be merged automatically, maybe you can update the PR? Sorry for the trouble.

LostPolygon:
Updated and resolved the conflicts, shoud be auto-mergable now.

LostPolygon:
Ping. It's auto-mergeable now, are there any problems with the code?

NathanSweet:
Thanks for following up. Sorry again for the delays. We've actually reviewed it, just waiting for @Fenrisul to try some things out. He promised he'd do it 2 days ago!

Other than that the only change I'd like is to rename `ExposedList` to `Array`. None of the names are ideal I know, but I already use the name "Array" in Java-land (which is also the reference Spine runtime, spine-libgdx), so it feels right at home for me and similarity between runtime ports is nice. You can do the rename if you like, else we can merge and then do it.

LostPolygon:
Great!
I can see why you'd like to rename `ExposedList` to `Array`, but I'm personally against it for similar reasons:
1) There is already a `System.Array` in .NET, which I personally use from time to time. Also, most .cs files have `using System;`, which means there'll be an ambiguous reference. Sure, it can be solved by doing `using Array = Spine.Array;` or something like that, but it feels dirty.
2) `ExposedList` is just a regular `List` with tiny modifications, why should it be totally renamed, losing all associations to `List`? If I see `WhateverList` in code, I already assume that it behaves like a regular `List`, no guessing required.

But yeah, nothing stops you from renaming it after the merge.

NathanSweet:
@LostPolygon It seems there is an issue where the triangles for a mesh with 4 vertices will be reused for a region. The triangles may not be in the same arrangement as the region, eg rendering it backward. This can also happen if two meshes have the same number of triangles but a different topology. I think the fix is to not reuse triangles for a mesh unless its the exact same mesh. Mitch and I are both slammed for some time, so feel free to jump on this grenade if you'd like! :)

I also noticed flip state is cached per attachment when it could be cached per bone (`attachmentsFlipStateTemp`).

LastState stores a number of items in separate lists. I wonder if it would make sense to store a single list whose entries are each an object containing all the pieces of data.

LostPolygon:
Ouch. I'll see if it's easy enough to fix, but there's chance I won't have time to work on this either :/

LostPolygon:
@NathanSweet
Is there any good reproduction skeleton/animation for this issue?

PR #377, opened 2015-03-03
Fixes uninitialized variable in flip timelines.

darionco:
Caused an exception when flip timelines were used.


NathanSweet:
Sorry for the delay. Flip timelines have been removed in Spine v3.

PR #379, opened 2015-03-04
[JS] Fixes frame count accessor in readAnimation

JaredSartin:
Improper frame count accessor was used - uses proper one.


daredevildave:
This should be merged.

badlogic:
This will be fixed in the spine-js runtime update to 3.4.0, see https://github.com/EsotericSoftware/spine-runtimes/issues/613

Thanks for your contribution!

PR #399, opened 2015-03-29
When mixing animation ignore invalid animator state.

charmke:
Mixing animation glitches fixed.


spine-unity/Assets/spine-unity/SkeletonAnimator.cs
```diff
@@ -105,6 +105,9 @@ public enum MixMode { AlwaysMix, MixNext, SpineStyle }
 			var stateInfo = animator.GetCurrentAnimatorStateInfo(i);
 			var nextStateInfo = animator.GetNextAnimatorStateInfo(i);
 
+			bool isValidStateInfo = (stateInfo.length > 0);
```
charmke: // better condition check
bool isValidStateInfo = (stateInfo.fullPathHash != 0);
bool isValidNextStateInfo = (nextStateInfo.fullPathHash != 0);

PR #403, opened 2015-04-02
[Unity] Fixed event syntax cause JIT compiling on iOS devices

yKimisaki:
Hello,

We use spine runtime as shared DLL in several Unity projects,
but event syntax cause JIT Compile Exception on iOS.
http://ja.esotericsoftware.com/forum/viewtopic.php?f=7&t=4204

Thanks.

### DIFF
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -30,6 +30,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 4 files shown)


pharan:
@yKimisaki This problem happens when you precompile Spine-C# and Spine-Unity into a DLL?

pharan:
I can't automatically merge this PR anymore but I saved the changes for future reference. Thanks for your contribution! We will keep this information noted.

PR #408, opened 2015-04-09
[Cocos2D-iPhone] fix transparent rendering

lpeancovschi:
Make transparent nodes to be hidden instead of rendered in white color


badlogic:
You can use `SkeletonRenderer::setOpacityModifyRGB` to change this behaviour. By default, the spine-cocos2d-objc runtime assumes that your atlas textures use premultiplied alpha.

PR #414, opened 2015-04-20
[as3] Fix strange behaviour of listeners

Fraggle:
Hello,

Just a couple of changes.
The change in add() will make sure a callback is never registered twice (slight change of behavior, consistent with the signals library, seems harmless to me).
The change in remove() will make sure that a listener is in the list before removing it (this one fix a bug, if indexOf == -1 then splice will remove the wrong listener).

Thanks for the hardwork !


Fraggle:
Can you merge that one please? I'm maintening my own clone just for this and the other bug I sent you.

badlogic:
Reasonable change, thanks! Sorry for it taking so long. I'm taking on the runtime business and will try to be more reactive :)

PR #415, opened 2015-04-22
[as3] add 'draw *MeshAttachment' support

zhongfq:


badlogic:
Thanks, we'll be incorporating this when fixing #614

PR #424, opened 2015-05-19
Qt conflicts because of variables named "slots"

ricardperez:
This will rename all the variables named "slots" to "slots_". This is because the library had to run in a Qt environment (using cocos2d-x as well). Qt uses a very powerful signal/slot system in its core, and for their projects the word slots is reserved.

### DIFF
--- a/spine-c/include/spine/Skeleton.h
+++ b/spine-c/include/spine/Skeleton.h
@@ -49,7 +49,7 @@ typedef struct spSkeleton {
 	spBone* const root;
 
 	int slotsCount;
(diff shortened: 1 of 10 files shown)


NathanSweet:
The QT conflict is unfortunate. I very much dislike using an underscore suffix in the public API. It seems you can compile with `-DQT_NO_KEYWORDS` so that `slots` does not get defined, then you can use `Q_SLOTS` (and `Q_EMIT`, `Q_SIGNALS`, etc).

PR #433, opened 2015-06-05
Fix for unexpected huge skeleton render scale when using SPINE_TK2D

invicticide:
I just recently switched from using the old spine-tk2d runtime to using the latest spine-unity runtime. This change caused all my skeletons to render at exactly 100x size. There was an explicit *100f at SkeletonDataAsset.cs:111 which doesn't seem to have a reason to be there (corroborated by [this thread](http://esotericsoftware.com/forum/spine-unity-ignore-tk2d-pixel-per-meter-setting-4233)) so I removed that, which fixed the in-game render size.

After making that change, the Inspector preview window for a SkeletonDataAsset stopped rendering if the **scale** property there was anything higher than about 0.18 (it defaults to 1). This doesn't make a lot of sense to me, but the condition at SkeletonDataAssetInspector.cs:651 did seem to be the culprit, so I changed that and it seems to work properly at all scales now (because... reasons? O_o)

Finally, the mouse wheel response in the Inspector preview got suuuuper touchy as the scale increased. That doesn't actually seem necessary since the renderer bounds don't seem to be changing with the scale (instead, the camera's orthographic size is changing), so I removed the scale multiplier at SkeletonDataAssetInspector.cs:860 and the mouse wheel response is much more consistent now.

I tested these changes against tk2dSpriteCollections using both Explicit and Pixels Per Meter sizing modes, at various sizes, and also against various skeleton scales as set in the SkeletonDataAssetInspector. Everything seems pretty solid, so... here you go!

### DIFF
--- a/spine-unity/Assets/spine-unity/Editor/SkeletonDataAssetInspector.cs
+++ b/spine-unity/Assets/spine-unity/Editor/SkeletonDataAssetInspector.cs
@@ -38,7 +38,7 @@ public class SkeletonDataAssetInspector : Editor {
 	private bool needToSerialize;
 
 	List<string> warnings = new List<string>();
-	
+
 	void OnEnable () {
 
 		SpineEditorUtilities.ConfirmInitialization();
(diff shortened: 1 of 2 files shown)


ArtFeel:
:+1: Please merge this PR, this is very important :warning:

ehudros:
+1

PR #442, opened 2015-07-07
Fix cocos2d multithreaded load bug

BrettThePark:
If a sprite is loaded from multiple threads at the same time it may cause the application to crash. The fix ensures that the atlas and skeleton file are only every read one at a time.

### DIFF
--- a/spine-cocos2d-iphone/3/src/spine/SkeletonRenderer.h
+++ b/spine-cocos2d-iphone/3/src/spine/SkeletonRenderer.h
@@ -39,6 +39,7 @@
 	bool _debugSlots;
 	bool _debugBones;
 	bool _premultipliedAlpha;
+  bool _skipVisibilityCheck;
 	ccBlendFunc _blendFunc;
 	CCDrawNode* _drawNode;
 	bool _ownsSkeletonData;
@@ -86,6 +87,7 @@
 @property (nonatomic, readonly) spSkeleton* skeleton;
 @property (nonatomic) bool debugSlots;
 @property (nonatomic) bool debugBones;
+@property (nonatomic) bool skipVisibilityCheck;
 @property (nonatomic) spBone* rootBone;
 
 @end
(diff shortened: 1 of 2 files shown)


badlogic:
I do not see how this prevents any race conditions. Loading the skeleton data in the `initWithFile` methods is essentially atomic. Can you elaborate how this code can exhibit multithreading issues?

BrettThePark:
If many skeletons are loaded on a background thread the code will often crash. A year since posting the PR, I don't recall exactly why.  The fix works by making sure that no two threads operate in the two `@syncronized` section at the same time, even if on different thread.

initWithFile works atomically if and only if it can be guaranteed to always be called on the same thread (generally the main thread).

Calling initWithFile in a background thread greatly increases the performance of loading sprites as they can be loaded in the background before they are needed using something like `dispatch_async`.

Skipping visibility checking can be helpful when we want thing stop be loaded out of screen in cases like background loading.

badlogic:
I understand the benefit of parallelizing the loading. However, the crash
would be contigent on data being shared between the loader threads. I don't
see which data would be shared. Can you give me a hint?

I'm happy to apply the fix, but I want to understand the reason for it
crashing first. ATM I'm jot able to reproduce it.

badlogic:
Digging into the code a little more, the only places i can see any sign of shared state access are in `_spAtlasPate_createTexture` which accesses the texture cache of the `Director` instance, and `_spUtil_readFile` which uses the `FileUtils` instance to read data. If those two classes aren't thread safe, then the thing may go boom. The Spine data structures aren't shared between threads. Maybe synching the implementations of `_spAtlasPage_createTexture` and `_spUtil_readFile` would make more sense?

BrettThePark:
I am not able to go back and alter the breaking code at the moment (however we are using the fixed code in production). From what I recall `_spAtlasPate_createTexture` and `_spUtil_readFile` do ring a bell with crashes. I believe in our case we do a background load and may call to create the sprite again on a different thread before it has finished loading on the original thread (causing either multiple reads to the same file at once or a read and write to the same cache spot for the Director).

badlogic:
Alright, that makes sense. Sorry for being an annoyance, just don't like spreading synch code throughout the codebase without understanding the underlying issue. Thanks for the input!

BrettThePark:
No problem. Glad you eventually got to it.

PR #449, opened 2015-07-18
In spine-canvas.js, slot's attachment's scaleX and scaleY not used…

chongdashu:
Fixed bug for spine-canvas.js where rendering did not account for slot's attachment's scaleX and scaleY properties.


chongdashu:
Closing for now, found a bug.

PR #451, opened 2015-07-18
[js] Fixed js-canvas renderer to account for scaleX and scaleY properties of attachments

chongdashu:
Previously a slot attachment's image with a modified Scale `x` or `y` transform value within Spine would not be reflected in the canvas renderer.

| Without Fix | With Fix |
|-----------------|------------|
| screen shot 2015-07-18 at 12 40 12 pm | screen shot 2015-07-18 at 12 40 06 pm  |


badlogic:
This has been fixed in the new spine-ts runtime.

PR #452, opened 2015-07-20
[Cocos2DX] Fixed render spine as 3D object.

fusijie:
Process flag to fix render spine as 3D object.


badlogic:
No longer applicable as we changed the rendering infrastructure in cocos2d-x

PR #454, opened 2015-07-23
Fixed error reported by address sanitizer 

Kabuto85:
```==13599==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000d770 at pc 0x605362 bp 0x7ffec4ab41b0 sp 0x7ffec4ab41a0
READ of size 8 at 0x60200000d770 thread T0
    #0 0x605361 in spSkeleton_create thirdparty/spine-c/source/common/Skeleton.c:67```


badlogic:
Thanks! Though this will only ever happen if a skeleton has 0 bones, which should be never :)

PR #465, opened 2015-09-02
[c] Add dynamic skin composition.

j-jorge:
Please merge this commit which adds a
`pSkin_createComposed(const char* name, struct spSkin** parts, int count)`
function allowing to create a skin by combining several other skins,
thus effectively allowing to build skins at run-time in the core part of
the Spine runtime.

This feature is an answer to requests like these:
- http://fr.esotericsoftware.com/forum/How-to-have-swappable-heads-and-skins-4304?p=20824
- http://fr.esotericsoftware.com/forum/How-do-I-implement-avatar-characters-4446
- http://fr.esotericsoftware.com/forum/mix-and-match-skin-attachments-618?p=17837

### DIFF
--- a/spine-c/include/spine/Attachment.h
+++ b/spine-c/include/spine/Attachment.h
@@ -55,6 +55,7 @@ typedef struct spAttachment {
 } spAttachment;
 
 void spAttachment_dispose (spAttachment* self);
+spAttachment* spAttachment_clone (const spAttachment* self);
 
 #ifdef SPINE_SHORT_NAMES
 typedef spAttachmentType AttachmentType;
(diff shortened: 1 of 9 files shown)


per-ludum:
We tested the code for this PR (after updating it for Spine 3.1) in our project and it works really well.

Dynamic skin composition should be a standard feature in the runtime.

I can post the updated code, if there is any use for it.

pharan:
Adding this to applicable topics: http://esotericsoftware.com/forum/Change-Region-Mesh-Path-at-Runtime-6689

badlogic:
We'll have to add this new API to all runtimes then. I guess @NathanSweet will want to have some input on this as well.

PR #466, opened 2015-09-04
[LUA] AnimationState trackCount bugfixes

reaver:
Fixes what I believe to be two bugs in the lua runtime AnimationState.

1) Adding an animation on an empty track will now play the animation. This is Line 214

2) The rest of the updates make trackCount hold count and not max track index as it previously did. This fixes an error where animations on the highest track index would stop if the second highest track index finished its animations and cleared the track.

Tested using modified SpineBoy example in main.lua in spine-love using the above mentioned use cases.

### DIFF
--- a/spine-lua/AnimationState.lua
+++ b/spine-lua/AnimationState.lua
@@ -211,7 +211,7 @@ function AnimationState.new (data)
 			end
 			last.next = entry
 		else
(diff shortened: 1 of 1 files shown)


NathanSweet:
Cheers!

PR #470, opened 2015-09-29
[lua][corona] Bug fixes for spine-corona

alexxwiz:
Hi!

I've found and fixed bug in spine-corona runtime with animating alpha value for attachment - it was not computed properly.
Also, there was a bug with setting color and alpha for display group - it can't be done in Corona, we need to go through group children and set values for them.


badlogic:
Thanks, the latest couple of spine-lua/corona iterations have all these things fixed!

PR #473, opened 2015-10-11
[Spine-Unity] Improved handling of Reimport All and some other weird asset import cases

LostPolygon:
1) In case user used `Assets -> Reimport All`, then during the import process, asset database is not initialized until some point. During that period, all attempts to load any assets using API (i.e. `AssetDatabase.LoadAssetAtPath`) will return `null`, and as result, assets won't be loaded even if they actually exist, which may lead to numerous importing errors. 

This situation also happens if Library folder is deleted from the project, which is a pretty common case, since when using version control systems, the Library folder must be excluded. So to avoid this, in case asset database is not available, we delay loading the assets until a bit later.

Unity *always* re-imports some internal assets after the process is done, so this method is always called once again in a state when asset database is available.

Checking whether `AssetDatabase` is initialized is done by attempting to load a known "marker" asset that should always be available. Failing to load this asset means that `AssetDatabase` is not initialized.

2) Also, sometimes for some weird reason Unity loses the internal Object pointer while reimporting the `SkeletonDataAsset`, and as a result, all comparisons with null returns true.
But the C# wrapper is still alive, so we can "restore" the object, by reloading it from its Instance ID.
This only happens occasionally on some random skeletons. Could not find any pattern, but this commit fixes this anyway.


pharan:
Merging these. You'll be around in case something's broken, right? Have you had any problems?

LostPolygon:
Sure. There shouldn't be any problems, and I haven't encountered any yet, so...

PR #475, opened 2015-10-28
Parsing and passing images path

jimon:
In some cases we need to have images path stored and passed to attachment loader. What is your opinion on it @NathanSweet ? Should it be implemented for all other runtimes as well ?

### DIFF
--- a/spine-c/include/spine/AttachmentLoader.h
+++ b/spine-c/include/spine/AttachmentLoader.h
@@ -57,7 +57,7 @@ void spAttachmentLoader_dispose (spAttachmentLoader* self);
 
 /* Returns 0 to not load an attachment. If 0 is returned and spAttachmentLoader.error1 is set, an error occurred. */
 spAttachment* spAttachmentLoader_newAttachment (spAttachmentLoader* self, spSkin* skin, spAttachmentType type, const char* name,
-		const char* path);
+		const char* path, const char* images);
 
 #ifdef SPINE_SHORT_NAMES
 typedef spAttachmentLoader AttachmentLoader;
--- a/spine-c/include/spine/SkeletonData.h
+++ b/spine-c/include/spine/SkeletonData.h
@@ -46,6 +46,7 @@ extern "C" {
 typedef struct spSkeletonData {
 	const char* version;
 	const char* hash;
+	const char* images;
 	float width, height;
 
 	int bonesCount;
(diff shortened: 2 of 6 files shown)


NathanSweet:
It's a little strange to need the `images` path, since that is usually only used by the editor. Spine tries to make it a relative path, but it can be an absolute path which is unlikely to work in an app. You could "subclass" the AttachmentLoader and have it know about what skeleton file it is loading, so it then knows where to look for images (or whatever you are doing with the `images` path).

jimon:
@NathanSweet the problem is that it's impossible to pass ```skeletonData``` at the moment, because it's created in ```spSkeletonJson_readSkeletonData``` and function ```spAttachmentLoader_newAttachment``` is called is called from it. So we don't have easy way of passing ```skeletonData``` to attachment loader. Another way could be is to parse json file again inside attachment loader, though it's not optimal.

Nice solution would be to pass ```skeletonData``` to ```spAttachmentLoader_newAttachment``` or just add pointer to ```skeletonData``` in ```spSkin```. Which one is more feasible ?

NathanSweet:
What do you need the skeleton data for?

You know what skeleton file you are about to load, so you could tell the subclassed AttachmentLoader which file before the JSON is even parsed. Probably the file name is enough for the AttachmentLoader to know where to find images (or whatever it is you are doing).

jimon:
I'm trying to get "images" field from json without parsing the file twice, this is needed because I'm using TexturePacker and want to use ["Prepend folder name" feature](https://www.codeandweb.com/texturepacker/documentation) so path to image in atlas will be the same as path to image on disk.

I see two ways to solve this :
- Modify spine model so images path will be "./", and then all attachments will point to "folder_name/image_name".
- Another way is to have images path "./folder_name" and attachments will point to "image_name". In this case I somehow need to get "./folder_name" string in my attachment loader :( Would be ideal to get it without modifying spine runtime, but I cant find a way to do it without loading json twice.

NathanSweet:
Extend AttachmentLoader and tell it what skeleton file you are loading before loading the skeleton file. In your AttachmentLoader you can prepend the name of that file to your atlas lookups. Probably your skeleton file name and your subfolder are the same.

It's a little easier to do your first bullet, and is what most people do. Spine has a search and replace feature.

I have mixed feelings about the proposed to change to pass more to the attachment loader. I have a feeling most people don't need this, and the `images` can be an absolute path.

PR #477, opened 2015-11-01
Restore blend function after drawing a skeleton on LibGDX

rafaskb:
When rendering slots with different BlendModes, the batch has it's blend functions modified, but they're never restored after the skeleton has been rendered. This commit fixes it by keeping track of it's previous functions and restoring them later.

### DIFF
--- a/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/SkeletonRenderer.java
+++ b/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/SkeletonRenderer.java
@@ -48,6 +48,8 @@ public class SkeletonRenderer {
 
 	@SuppressWarnings("null")
 	public void draw (PolygonSpriteBatch batch, Skeleton skeleton) {
+		int oldBlendSrc = batch.getBlendSrcFunc();
+		int oldBlendDst = batch.getBlendDstFunc();
 		boolean premultipliedAlpha = this.premultipliedAlpha;
 		BlendMode blendMode = null;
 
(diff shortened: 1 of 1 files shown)


badlogic:
It's a common libGDX pattern to have users set up the batch state before drawing and not reset the batch state to what it was. As such, I'm afraid we won't merge this PR. Thanks for the contribution though!

rafaskb:
That does make sense, but without modifying Skeleton Renderer's code directly, it's impossible to restore the previous function between two skeletons that are attached to each other.

Rendering example

In the example above there are three models being rendered, the player, the weapon and the yellow gadget. Both equipments are programmatically attached to the player hands using SkeletonAttachment, and when the parent is rendered all attachments are directly rendered as well. The weapon model makes use of an additive blending function in one of the slots, which causes the player's model to be rendered incorrectly.

rafaskb:
Just tested with Spine 3.4.02 and the problem persists.

badlogic:
I see your point. I'll let @NathanSweet call it. Only the gdx runtime can deal with SkeletonAttachments iirc, and its used in the editor as well. Worst case, you either have to subclass the renderer or simply modify your copy of the source.

NathanSweet:
The next skeleton to draw may use the same blend mode last thing drawn be the previous skeleton, in which case restoring the old blend func would cause an unnecessary batch flush. The real problem is we're being overly clever. Incoming commit!

rafaskb:
That solves the problem, thank you 😄

PR #480, opened 2015-11-19
[Cocos2d-x] Add `gotoAndPlay` `gotoAndStop` `pauseAnimation` and `resumeAnimation` for SkeletonAnimation.

iWoz:
Add some common used APIs for `SkeletonAnimation` used in Cocos2d-x.

* `void SkeletonAnimation::gotoAndPlay(int frame)`: jump to frame and continue play animation.
* `void SkeletonAnimation::gotoAndStop(int frame)`: jump to frame and pause animation.
* `void SkeletonAnimation::pauseAnimation`: pause animation.
* `void SkeletonAnimation::resumeAnimation`: resume animation.


badlogic:
Thanks for the contribution. We are reworking AnimationState (see https://github.com/EsotericSoftware/spine-runtimes/issues/621) which will most likely cover your improvements as well in some form.

PR #486, opened 2015-11-30
[SFML] Fixed include statments

tsteinholz:
* Removed include statements from the cpp
* Removed duplicate includes
* Added include statements from the cpp to the header


tsteinholz:
Hello?

PR #490, opened 2015-12-05
[Spine-Unity] Fix for Frame Jump Bug when mix time resolves to zero

mattluard:
Originally reported [here on the forums](http://esotericsoftware.com/forum/Unity-One-frame-of-wrong-animation-4719), couldn't find a Github Issue open for the problem.

The graphical bug is this; for the very last frame of a mix between animations, the mix `time` will resolve to zero, and the renderer will display the first frame of the animation. It's a very obvious bug. Imagine half way through an animation, suddenly rendering the first frame. Simply skipping that update, the solution I'm proposing here, cleans the problem up for me.

My first contribution to this repository, very happy to rework things if not up to scratch.

### DIFF
--- a/spine-unity/Assets/spine-unity/SkeletonAnimator.cs
+++ b/spine-unity/Assets/spine-unity/SkeletonAnimator.cs
@@ -108,6 +108,7 @@ public class SkeletonAnimator : SkeletonRenderer, ISkeletonAnimation {
 						continue;
 
 					float time = stateInfo.normalizedTime * info.clip.length;
+					if (time == 0) continue;
 					animationTable[GetAnimationClipNameHashCode(info.clip)].Mix(skeleton, Mathf.Max(0, time - deltaTime), time, stateInfo.loop, null, weight);
 				}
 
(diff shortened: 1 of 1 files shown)


LostPolygon:
Hm... I've actually made a PR for the exact same bug, but I fixed it in a bit different way:
https://github.com/EsotericSoftware/spine-runtimes/pull/482
It's not actually `AnimatorStateInfo.time` becoming 0, it's the whole `AnimatorStateInfo` being empty for that one frame. i guess relying of `time == 0` isn't as good of idea, as `time == 0` is actually a perfectly normal situation.

mattluard:
Yes, time should be allowed to be zero without displaying the wrong frame, so it's probably true that I'm fixing the symptom rather than the deeper underlying issue.

pharan:
@LostPolygon Already merged the other one. Should this one be closed?

LostPolygon:
@pharan I think so. My PR _should_ be fixing the same exact issue, but in a better way.

PR #491, opened 2015-12-05
[LUA] Change null to nil

ficoos:
nil is the actual keyword. 'null' works in non strict mode as all undeclared variables are considered 'nil' in non strict mode.


NathanSweet:
Cheers!

PR #510, opened 2016-01-22
Update spine.js

wyklion:
bug.


badlogic:
This will be part of the 3.4.0 update of spine-js, see https://github.com/EsotericSoftware/spine-runtimes/issues/613

PR #514, opened 2016-02-11
[Unity] Disallow multiple SkeletonRenderers.

pharan:
Disallow multiple SkeletonRenderers on one GameObject. This is unsupported, makes no sense, and also causes Unity to go nuts.


pharan:
Applied to v3

PR #519, opened 2016-02-16
Fixes in spine-csharp and spine-c

Hinidu:
Hi again, I've merged my own ports with yours and found that commits https://github.com/EsotericSoftware/spine-runtimes/commit/b6f7e50ff79d280ba4a8958c4a07e04b920b5957, https://github.com/EsotericSoftware/spine-runtimes/commit/c7bb3f3af875608bf3d02c80dd1d852f5fde95c9 and https://github.com/EsotericSoftware/spine-runtimes/commit/2650e2679d5992b98f00a63fc270f186791a14a0 were not ported yet and some issues like not freeing `updateCache` in `spSkeleton_dispose`.

### DIFF
--- a/spine-c/include/spine/Animation.h
+++ b/spine-c/include/spine/Animation.h
@@ -296,7 +296,7 @@ typedef struct spEventTimeline {
 
 spEventTimeline* spEventTimeline_create (int framesCount);
 
-void spEventTimeline_setFrame (spEventTimeline* self, int frameIndex, float time, spEvent* event);
+void spEventTimeline_setFrame (spEventTimeline* self, int frameIndex, spEvent* event);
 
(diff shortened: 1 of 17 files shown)


pharan:
The changes to AnimationState and its accompanying changes aren't done yet. I don't think they're ready to be applied to the other runtimes.

Hinidu:
Ok, but what about other things? For example memory leak in `Skeleton.c`. I can extract parts of this commit which you think can be useful.

NathanSweet:
Leak has been fixed, thanks. I've also updated most of the missed porting you pointed our with your PR. It's great to have multiple eyes on the runtimes! Cheers!

Hinidu:
Thank you @NathanSweet

PR #524, opened 2016-02-20
Added requires in attachment loader

ggcrunchy:
Minor fixes for off-by-one indexing and loop counts for skinned meshes

Some fixes for names or counts, again for skinned meshes

(This does not include the mesh logic itself, nor have I tested basic meshes.)


NathanSweet:
Thanks! Would you be so kind as to sign and email (contact@estoericsoftware.com) our CLA?
http://esotericsoftware.com/files/cla.txt
Cheers!

jonbro:
oh nice! this should solve the issue that I was having.

PR #529, opened 2016-02-26
Screen shader for color blending

iamsed:
Hello, this is a Screen shader (based on the Screen effect from the photoshop) for the Unity runtime.


pharan:
I don't think we'll be adding specialty shaders to the official runtime we have to maintain.
But we'll link to this code in the forums if you leave it somewhere. If not, I can just add it there as an attachment.

PR #536, opened 2016-03-12
Improved material overrides

LostPolygon:
Added ability to override materials per-material (useful if you want to change the whole material without messing with any specific slots), and a `SkeletonRendererMaterialOverride` `MonoBehaviour` to allow setting up overrides from inspector.


pharan:
@LostPolygon I just applied all the new stuff (including namespaces) but I can't merge this anymore. If you want to make another PR, I'll merge it.

Or I can make the edits to SkeletonRenderer and you can PR the module.
Or I can just make all the edits, since the module is pretty simple too.
Up to you.

LostPolygon:
@pharan 
I guess it'd be easiest for you to fix, since, honestly, after your new stuff I currently have no idea how to fix it myself :)

pharan:
Okay, I'll make the changes.

The current system is actually mostly the same. I just removed a bunch of old stuff that didn't make sense anymore (they came from way back when SkeletonRenderer was first implemented).

And the new SmartMesh class is actually just a coupling of the old mesh1/mesh2 and your MeshState system. It seemed like an obvious way to go. Doing it that way also removed the need for all those hacks we had to add to clear both meshes and states and stuff like that.

pharan:
The SkeletonRenderer modification is done: https://github.com/EsotericSoftware/spine-runtimes/commit/1589d147f966c5d74be34734773bcac875ffd8fd

I did the material replacement at the submesh level and not in the per-slot for-loop. That way, we end up with only as many dictionary lookups as there are submeshes. So 1 submesh should mean 1 lookup instead of n-number of slots.

I also named the dictionary property `CustomMaterialOverride`.
You can rename that in your script and it should work exactly the same (theoretically).

LostPolygon:
@pharan 
I've did the changes on my side. It pretty much worked after just renaming that property.

LostPolygon:
@pharan 
Merge time?

pharan:
Yeah, in a bit. I'll also likely move this into the Spine.Unity.Modules namespace.

pharan:
Can you describe the use case (and how to use) for the `SkeletonRendererMaterialOverride` script?

It's not immediately clear to me. I mean it looks like the override items can be enabled and disabled piecemeal but changes there doesn't automatically apply it, right? And is it designed to only have one of that script at a time per SkeletonRenderer? Will it break if there's two?

(You can do this in the forum if you want, so people understand how to use it.)

LostPolygon:
@pharan 
Primarily, `SkeletonRendererMaterialOverride` is designed for setting custom materials from the inspector. I, personally, rarely have the need to change materials in runtime, but I often need different skeleton instances to have different materials set to them during scene editing. You *can* use it from the code, but I'm not sure why would you want that if you can just use `SkeletonRenderer.CustomMaterialOverride` and `SkeletonRenderer.CustomSlotMaterials` directly.

It won't "break" if there are multiple instances of it per SkeletonRenderer, as long as overrides do not overlap. I mean, if some code overrides some material using `SkeletonRenderer.CustomMaterialOverride` to one value, and some other code overrides it to some other value, unavoidably only the last one will actually work. But as long as overrides do not overlap at any given moment - it will work, and there are some safeguards to ensure that.

Changes are not applied automatically, since there is no good way to listen to changes in `List<T>`, and I didn't wanted to overcomplicate it. After making changes to the override, you have to actually apply them. I guess a method like `ReapplyOverrides` will be useful for this, though. Something like this:
```csharp
public void ReapplyOverrides() {
    RemoveCustomMaterialOverrides();
    RemoveCustomSlotMaterials();
    SetCustomMaterialOverrides();
    SetCustomSlotMaterials();
}
```

pharan:
I see. So it's mostly a an inspector and editor tool and an object to store serialized overrides and apply them when the scene starts.

The public properties and methods are just in case?

LostPolygon:
Yep, correct.

pharan:
I'm going to change the access qualifiers of a bunch of things to match the inspector-only expectation. Don't want users to think it's supposed to be used for code.

pharan:
Problem with the inspector. Changing the slotName causes the script to reapply the override lists, but fails to remove changes prior to the slotName change.

LostPolygon:
I'm not sure it's inspector-**only**. Sure, it's main purpose is to allow setting the materials in the editor, but what if I also need to dynamically override some slot in other code? `SkeletonRendererCustomMaterials` shouldn't be a black box without any means to change the overrides.

LostPolygon:
Huh, you can change the slot name? Why would you want to do that?..

pharan:
In case you're still picking the correct slot and misclicked. Or you're trying to find which slot it was 'cause you're not sure. So you'd pick a slot name, realize it was wrong, then pick another one.

LostPolygon:
@pharan 
Ah, I see. Well, it can be fixed by making the inspector track the changes, but that will require making `Remove/SetCustomMaterialOverrides()` and `Remove/SetCustomSlotMaterials()` public again.

PR #539, opened 2016-03-17
Update MonoGame iOS example in line with recent changes.

RandolphBurt:
* Migrate to unified API to support 64 bit platforms.
 * Remove call to game.exit() in ExampleGame.cs as this is no longer supported.
 * Use MonoGame.ios Nuget package instead of relying on external repo being present on the machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,9 +42,11 @@ spine-csharp/src/*.cs.meta
 
 spine-monogame/xamarinstudio-ios/src/bin
 spine-monogame/xamarinstudio-ios/src/obj
+spine-monogame/xamarinstudio-ios/src/packages
 
 spine-monogame/xamarinstudio-ios/example/bin
 spine-monogame/xamarinstudio-ios/example/obj
+spine-monogame/xamarinstudio-ios/example/packages
 
 spine-monogame/windows8-store/src/bin
 spine-monogame/windows8-store/src/obj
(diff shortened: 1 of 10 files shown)


pharan:
Thanks. So what's supposed to be called instead of Game.Exit?

RandolphBurt:
Good point!
As it is an iOS app then so you don't need to close it.  Looking in the MonoGame source code I can see it has the following attribute which forces a compile time error if you try to call game.exit() in an iOS or windows store app....

_#if IOS || WINDOWS_STOREAPP && !WINDOWS_PHONE81
        [Obsolete("This platform's policy does not allow programmatically closing.", true)]
#endif
_

Therefore I would suggest the correct fix is not for me to delete the existing code, but instead change the exampleGame.cs code to apply the same logic. i.e. to be as follows instead...

_#if (!WINDOWS_STOREAPP || WINDOWS_PHONE81) && !IOS	
			if (GamePad.GetState(PlayerIndex.One).Buttons.Back == ButtonState.Pressed)
					this.Exit();
#endif
_

If you are happy then I'll add that to the pull request. (Or I can squash the commits and do a new pull request - whichever you prefer). 
Cheers

pharan:
Yeah, adding the IOS symbol to the `#if` makes sense. Go for it.

RandolphBurt:
OK. I've made the change.  Let me know if any issues/problems.
With regards to the MonoTouch API reference, I believe it is just a name change.

PR #540, opened 2016-03-25
renderer should support normal blend mode

jareguo:
support `SP_BLEND_MODE_NORMAL` in SkeletonRenderer

Test animation：
[aitiwen.zip](https://github.com/cocos2d/cocos2d-x/files/189123/aitiwen.zip)

Wrong rendered result: (a spot light appeared on "COCOS2D")
image

Fixed result: (the spot light become transparent)
image


NathanSweet:
Fixed in latest spine-cocos2dx (renderer has been rewritten).

PR #551, opened 2016-04-11
[js] added bower.json

goldsam:


NathanSweet:
Is there a way for Bower to work without us adding a file? I don't like that the version number is in the file, it will quickly become out of date.

goldsam:
You can omit the version. The current [bower.json specificaiton](https://github.com/bower/spec/blob/master/json.md#version) had deprecated the version property Bower now uses git tags instead :)

NathanSweet:
Is there a way for Bower to work without us adding a file?

goldsam:
No. The bower file references the dependency artifact (./spine-js/spine.js), and enabled build tools (e.g. grunt and gulp plugins) to consume it during a build.

badlogic:
We have replaced the spine-js runtime with spine-ts (TypeScript). This PR no longer applies.

PR #552, opened 2016-04-13
build success with vs2015 on win7

yangshuang-developer:
libSpine is missing if somebody want to run the example with visual studio do as the readme.RD file


pharan:
What is libSpine and why would it be part of the generic spine-c vcxproj?

badlogic:
spine-c and child runtimes have a new build system. Should work as per README.md.

PR #553, opened 2016-04-14
SkeletonRendererCustomMaterials: correctly reapply changes when slot name changes

LostPolygon:
@pharan 
Fix to incorrect behavior when editing the slot material override list in the inspector (#536).
It's not the cleanest solution, but oh well, I guess `Remove/SetCustomMaterialOverrides()` and `Remove/SetCustomSlotMaterials()` have to remain public.


pharan:
I'll stick this in a branch first. https://github.com/EsotericSoftware/spine-runtimes/tree/unity-custommaterials

Already merged there. Will do further poking at it in that branch.

PR #568, opened 2016-05-11
fit spine 3.2

shinjiyu:
this return sentence   makes this function does not work well with json extension.


NathanSweet:
spine-c [does not support 3.2 yet](https://github.com/EsotericSoftware/spine-runtimes/tree/master/spine-c#spine-version). Ignoring a timeline (I assume you use a shear timeline, which is only in 3.2) is not a good idea. You can use 3.1.08 until spine-c is updated.

PR #569, opened 2016-05-11
[c] Prevent crash when setting new animation during event callback on same trackId

shinjiyu:
user may set new animation with the same trackid  in  callback.
so do all clean jobs before callback to user's code


NathanSweet:
AnimationState has been upgraded starting with Spine 3.5. This and many other issues have been fixed, see #621.

PR #574, opened 2016-05-12
[Unity] Custom Materials serialization cleanup.

pharan:


LostPolygon:
@pharan 
`ReapplyOverrides` method should remain there as a public method. Otherwise there is now no way to modify the overrides. There's no purpose in having `CustomSlotMaterials` and `CustomMaterialOverrides` public properties otherwise.

One small thing I've just thought: `SlotMaterialOverride` and `AtlasMaterialOverride` should probably be made classes instead of structs, to allow just referencing the class instance and modifying it directly from other code. Could be handy.

pharan:
I'll just remove those properties then.

LostPolygon:
@pharan 
Uh, wait, why?.. Even though it's not designed to be used from code, I don't think it should be a black box with no way to access it or change anything. Just bring back `ReapplyOverrides` public method and it'd be complete.

pharan:
It would have the same weird mismatch behavior that the old inspector had whenever you change it and reapply the overrides.

You'd have to do the same caching thing the inspector now, but at runtime, which would be overkill and unnecessary compared to if they just managed SkeletonRenderer directly.

It's not a black box if it's source-available. People can modify or adapt the code however they want. And it'll work better in code that way.

LostPolygon:
Okay then.

PR #600, opened 2016-06-03
Fixing bugs in Lua runtime; adding support for meshes in Corona runtime; adding spineboy-mesh demo

### LINKED ISSUE #339: Mesh doesn't work with Corona
Hi everyone. 
 I try to use Spine with Mesh in Corona but it doesn't work.  I try to fix some small bugs like : forgeting require MeshAttachment in AttachmentLoader.lua, calling SkeletonData.findSkin instead of SkeletonData:findSkin in SkeletonJson.lua, or something like that.
 But it has some huge bugs with array index, attachment verticles, v.v.v, in SkeletonJson.lua. It seems that the one who ported it from other languages didn't remember to change start index from 0 to 1. And in Lua, table doesn't have length method, it requires using #table to get its number of children.
 I tried to fix as many errors as I can. But function: 
-------------------------------
...
 local ffd = map["ffd"]
		if ffd then 
...
-------------------------------
in Skeleton.lua has too many bugs with Lua which I don't have enough experience in Spine to fix them. Please take a look at this.
 Thanks for reading.

Shchvova:
Hello! I was working on implementing meshes with Corona's new [display.newMesh()](https://docs.coronalabs.com/daily/api/library/display/newMesh.html). It appears to be fairly easy to implement, which I did. It resolves #339.

While doing it I encountered couple bugs in Animation.lua. I don't think Meshes part of Lua framework was ever used, so here's bugs I fixed:

* ffd animation was ignoring last frame. Lua has 1-based arrays, so last frame is `frames[#frames]`, not `frames[#frames - 1]`
* nil exception when accessing array lenght after it's confirmed nil. First we check if `not vertices` and in next statement we're trying to check for it's length, which causes exception. This code is logically equivalent to code before but avoids checking potentially nil array for length.


NathanSweet:
CLA received, thank you!

It looks good. It would be nice to render region attachments with meshes, as they can be deformed into rhomboids. However, maybe we should wait until indexed meshes are batched? Otherwise we could convert it into a triangle strip (or Corona could do that internally?).

spine-corona can render meshes everybody! Rejoice! :D

Shchvova:
Welcome! Glad to make Corona & Spine customers happier. Unfortunately Corona doesn't convert indexed triangles to triangle strips automatically, but it wouldn't be a problem to make meshes explicitly with triangle strip with `display.newMesh{mode='strip', ... }`. But this is another story. Also, You can make rhomboids out of regular Corona sprites, with (path.x,y)[https://docs.coronalabs.com/guide/graphics/3D.html] properties, there is not even need for mesh objects here. But this is another story.

NathanSweet:
May be a little tricky to determine vertex offsets, but possible. Meshes would be easier, and I expect we'll find a solution for batching at some point. It it possible to do non-perspective scaling with the `path` property?

Shchvova:
Basically you're getting control over vertices in a quad with `rect.path.x1` ... `rect.path.y4` properties. It can be little tricky to calculate exact offsets you'll need to set.
Also, it's OK to use mesh, and it would batch OK if you'll do something like
`display.newMesh({ mode = "strip", vertices = { 0,0, 0,100, 100,0, 100,100 } })`

Shchvova:
By the way. Thinking of immediate things we can do to make Corona + Spine better is to implement atlas loading in Lua/Corona. I plan to port [this](https://github.com/EsotericSoftware/spine-runtimes/blob/master/spine-csharp/src/Atlas.cs#L98) to Lua and Corona's Image Sheets on my spare time...

NathanSweet:
Agreed, having an atlas with named regions in Corona would be great!

Shchvova:
So, I wrote simple exported of Spine Atlases. It doesn't parse "rotate" properly yet, but it's not a problem.

I tried to use Raptor animation you suggested, and it was total failure. I found multiple issues inside `spine-lua` runtime, most of which I didn't fix yet. It seems that Raptor uses little more features than I had in my spineboy mesh animation. Mostly like faulty array boundaries and 0-based indices...
I wouldn't recommend to publicize Mesh support because it seems there are still things to troubleshoot in `spine-lua`.

NathanSweet:
Sorry, it's likely there are problems since we've not yet had a Lua game toolkit that renders Spine meshes. We'll be updating the runtimes over the next couple weeks and will be sure to fix those bugs.

Shchvova:
Not a problem. It just examples I was able to run before trying to use atlases worked fine.

SimonFearby:
Thanks for the work to date (lowly corona developer here making business and augmented reality apps) but watching this work to open up possibilities on my next app.

PR #605, opened 2016-06-07
[cocos2dx] When the Director is paused while spine animations are running it crashes the app

gitshah:
I noticed that the scheduled update calls do not get called when Director is paused but draw callback continue to be called. This was causing the app to crash.

The solution is really simple, instead of doing the _position reset in update callback, I registered a cusom event listener for "director_after_draw" event. This event is raised after every update call irrespective of whether the director is paused or not.

Tested it in our game and it seems to be fixing the crash.


badlogic:
Thanks, merged manually.

gitshah:
Awesome thanks!

PR #610, opened 2016-06-15
[Cocos2dx] fixed SkeletonRenderer creation with given atlas (was crashing)

clegomane:


badlogic:
Thanks, merged manually.

PR #616, opened 2016-06-22
[spine-unity] Support sprite attachments with pivot placed outside the sprite

gloomybrain:
The case here is as follows:
 - create random unity sprite prefab
 - place sprite's pivot point outside of the sprite (say, pivot U is 2.53 and pivot V is -1.22)
 - try to attach this sprite via the SpriteAttacher script

Expected result here would be seeing the sprite sticking with it's pivot point to the placeholder.
Actual result here looks like the sprite's pivot point is somewhere in U: 1 and V: 0.

The bug is caused by Mathf.InverseLerp - it just cannot return values outside of [0; 1] range.

So the solution would be to make our own version of InverseLerp, which handles such situations correctly.


pharan:
Thanks for this contribution! Just at a glance, it looks ok.
I'll review it and we'll merge in a few days if things check out.

gloomybrain:
Thanks a lot for merging, keep up the good job =)

PR #620, opened 2016-06-23
[lua][corona] Initial Lua and Corona Atlas support

Shchvova:
* [Lua] Adding Spine atlas parser
* [Corona] Adding ability to load atlas files
* [Corona] Added example of using atlas

# Documentation:

## Lua runtime: `spine.Atlas.parse`

`Atlas.lua` introduces `spine.Atlas.parse(atlasPath, baseDir)` funciton. This function can be use across Lua runtimes to implement their their own image sheet bindings. It will return Array of `Page`s or `nil` if error.

### `Page` properties:
* `name` - filename of the page (`"raptor.png"`)
* `size` - dimmensions of original image (`[1022,1022]`)
* `format` - texture format (`"RGBA8888"`)
* `filter` - filter for texture scalin up and down (`["Linear","Linear"]`)
* `wrap` - wrap modes (`"none"`)
* `regions` - Array of `Region`s on this page

### `Region` properties:
* `name` - region name (`"torso"`)
* `rotate` - is image rotated in texture (`true`)
* `xy` - image frame position on texture (`[610,647]`)
* `size` - image frame size on texture (`[54,91]`)
* `splits` - (optional)
* `pad` - (optional)
* `orig` - original image size (`[54,91]`)
* `offset` - offset if image was trimmed (`[0,0]`)
* `index` - `-1`

## Corona runtime: `spine.GetAtlasSprites`

`spine.lua` introduces `spine.GetAtlasSprites(atlasName, baseDir)`, which returns `Sprites` table. This funciton would parse atlas, locate atlas textures, load them to Corona's image sheets and create convinience bindings and functions. User can use convenience functions, or write own wrappers.

#### Convenience functions:
* `ATLAS_HELPER_createImage` - can be assigned to `skeleton.createImage`
* `ATLAS_HELPER_createMesh` - can be assigned to `skeleton.createMesh`
* `ATLAS_HELPER_setupSkeleton` - use this on skeleton to assing both creation functions (`sprites.setupSkeleton(skeleton)`)

### Region name bindings:
All other elements of `Sprites` table would contain binding `sprites["region_name"] = sheetReference`, where each `sheetReference` contains following elements:

* `frame` - number of frame in image sheet (`42`)
* `region` - reference to `Region` table
* `page` -  reference to `Page` table. `Page` table would also get new property - `sheet`, reference to Corona's image sheet where sprites can be accessed.

As result, to create image for `attachment` code would look something like `display.newImage( sprites[attachment.name].page.sheet, sprites[attachment.name].frame )`. For more rubust example see `ATLAS_HELPER_createImage` in `spine-corona/spine.lua`.

# Known Issues:

* Corona would not work with padded/cropped atlases.
* Possible (but unprobable) name collision with region names and `ATLAS_HELPER_*` functions.


Shchvova:
You already have my signed CLA.

badlogic:
Awesome, thanks!

PR #619, opened 2016-06-23
Update spine-c to v3.3 (path feature is not implemented yet).

Hinidu:
Hello!
I've started to port changes from `spine-libgdx` to `spine-c` one day before you've merged support for Spine 3.2. Btw the same thing happened for Spine 3.0 update, and I find these coincidences very funny. :smile:
Though I've tried to port the code for Spine 3.3 too (except `Path` because we don't use it yet).
I didn't test it yet because we should update our custom code to run it with new runtime and if I will find some bugs I will update this PR.
It would be great if you will find my work useful and it would be awesome if you will provide some feedback so I will be able to improve it.
Thank you in advance.
PS: I've changed IkConstraint.c to tabs (in the last commit it was converted to spaces, I think it was accidentally) so diff will be prettier if you will put `?w=1` at the end of url in your browser.

PR #624, opened 2016-06-30
[spine-c] fixes with worldVerticesLength usage.

Hinidu:
Hi!
We've encountered some issues with weighted meshes and I found some mistakes made in #619.
We still didn't get the stable version unfortunately :disappointed: 
I will push my ongoing changes here.

### DIFF
--- a/spine-c/src/spine/MeshAttachment.c
+++ b/spine-c/src/spine/MeshAttachment.c
@@ -82,6 +82,8 @@ void spMeshAttachment_computeWorldVertices (spMeshAttachment* self, spSlot* slot
 void spMeshAttachment_setParentMesh (spMeshAttachment* self, spMeshAttachment* parentMesh) {
 	CONST_CAST(spMeshAttachment*, self->parentMesh) = parentMesh;
 	if (parentMesh) {
+		self->super.worldVerticesLength = parentMesh->super.worldVerticesLength;
+
 		self->super.bones = parentMesh->super.bones;
 		self->super.bonesCount = parentMesh->super.bonesCount;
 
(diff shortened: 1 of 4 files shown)


badlogic:
What do you mean with "you didn't get the stable version"? I'm currently fixing up the remainder of the spine-c runtime. Thanks for your fixes!

Hinidu:
I mean that our game still crashes (even with these fixes) on reading of the skeleton with the weighted mesh.

badlogic:
Oh, could you maybe send us an example spine project/export? Our weighted mesh examples work as intended.

Hinidu:
We fixed some usages of `worldVerticesLength` in our custom code and it doesn't crash now but mesh rendering is broken though.
> Oh, could you maybe send us an example spine project/export? Our weighted mesh examples work as intended.

I can't do that at least now. Perhaps we will just wait your release and sync our code with your again.

badlogic:
You could try to run the goblin or raptor examples in your modified
runtime. If those work for you, then you found a bug not covered by our
changes yet.

Hinidu:
We fixed that! :sunglasses: Look at https://github.com/EsotericSoftware/spine-runtimes/pull/624/commits/274f20102e9f0959596e6b6667d992a1a46e1a92 I will make a PR for spine-csharp (we encountered the same problem in this runtime and now it is fixed) and spine-libgdx (I think it works the same way) in a minute or two.

badlogic:
Awesome! I actually just stumbled across this weird side effect heavy code when enabling pedantic warnings in Clang. @NathanSweet is to smart for his own good! :D

NathanSweet:
Silly C languages!

PR #625, opened 2016-07-01
[csharp][libgdx] fix iterating bones in VertexAttachment.ComputeWorldVertices.

Hinidu:
Postincrement affects its variable only when entire expression is evaluated.
The same fix for spine-c is here: https://github.com/EsotericSoftware/spine-runtimes/pull/624#issuecomment-229898936


badlogic:
Thanks!

badlogic:
Fixed it in spine-c as well.

PR #627, opened 2016-07-05
[cocos2d-x] Fix implicit conversion warning

mogemimi:
This PR fixes the following warning when compiling `spine-cocos2dx.cpp` which included in cocos2d-x with Xcode 7.3.1 and Clang:

```
spine-cocos2dx.cpp:52:12: implicit conversion loses integer precision: 'ssize_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
```


badlogic:
Thanks! Applied manually due to merge conflict.

mogemimi:
Thank you for applying the patch! :thumbsup:

PR #628, opened 2016-07-05
[c], [sfml], [cocos2dx], [cocos2d-objc] updated to 3.3.07

badlogic:
This update adds support for the new path feature in spine. Weighted mesh attachments have been removed and merged into the standard mesh attachment. You will have to export your Spine models from Spine 3.3.07 to use them with the updated runtimes.


badlogic:
Closed in favor of rebased version https://github.com/EsotericSoftware/spine-runtimes/pull/629

PR #635, opened 2016-07-14
[Unity] Added BoundingBoxFollower.isTrigger

Xelnath:
Added the IsTrigger property to BoundingBoxFollower and BoundingBoxFollower editor.


pharan:
Thanks, Xelnath! Looks good! I'll merge it in a while.

PR #636, opened 2016-07-14
[spine-c] Fixed spBone_worldToLocalRotationY

iboB:
* crashes when bone has no parent
* doesn't calculate local rotation when bone has parent

I stumbled upon this while porting 3.4 to spinecpp :)


badlogic:
Good catch, thanks!

PR #643, opened 2016-07-19
[spine-c] Fixed calculation of number of path constraint timelines.

iboB:
* The json structure is that the path constraint timelines are added per path constraint
* Instead of the number of timelines what was calculated was the number of path constraints for which we have timelines
* If there were more timelines than path constraints, this led to memory corruption (writing past the buffer)


badlogic:
Ugh, silly me. Thanks!

PR #644, opened 2016-07-19
[spine-c] Fixed uninitialized value

iboB:
* First space of path constraint spaces remained uninitialized
* This led to various visual bugs with path constraints


badlogic:
Hm, what does your path constraint look like? Thansk!

iboB:
You can see the fixed bugs with both the tank and the vine example meshes.

PR #650, opened 2016-07-28
#Use Data::takeBuffer replace deep copy data.

halx99:
Avoid unnecessary memory alloc & copy

### DIFF
--- a/spine-cocos2dx/src/spine/spine-cocos2dx.cpp
+++ b/spine-cocos2dx/src/spine/spine-cocos2dx.cpp
@@ -75,11 +75,8 @@ void _spAtlasPage_disposeTexture (spAtlasPage* self) {
 }
 
 char* _spUtil_readFile (const char* path, int* length) {
-	Data data = FileUtils::getInstance()->getDataFromFile(
-			FileUtils::getInstance()->fullPathForFilename(path).c_str());
+    Data data = FileUtils::getInstance()->getDataFromFile(
+			FileUtils::getInstance()->fullPathForFilename(path));
     if (data.isNull()) return 0;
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks!

badlogic:
Looks like that method on CCData got removed in the latest release... We'll have to revert this PR.

Does cocos2d-x have any version policy? Can minor releases break API?

halx99:
The 3.13 have this API

ricardoquesada:
the issue with this PR is that is doing an "upcast" from `int` to `ssize_t` and might overwrite the stack in certain 64-bit platforms where `sizeof(ssize_t` is bigger than `sizeof(int)`

badlogic:
Jupp, already reverted it.

ricardoquesada:
@badlogic sent a new version of the PR that avoid the stack overflow

badlogic:
We have a couple of users on < 3.12, so we'll keep the copying version for
a little while. Thanks for the PR!

ricardoquesada:
makes sense... I think `takeData` was added recently.

halx99:
@bodlogic Currently, I think you can use the othet compatiable API: getFileData, enen though it's marked as  depecrated

ricardoquesada:
@badlogic what if we add something like this?
```c++
#if COCOS2D_VERSION >= 0x00031200 
    // new code
#else
     // old code
#endif
```

would yo accept a patch like that? thanks

badlogic:
Yeah, we can do that. I just worry that eventually the whole code base is
littered with #ifdefs. I'll merge this PR and add the #ifdef manually.
Thanks!

PR #652, opened 2016-07-28
Improved performances when applying color and alpha on vertices

Julienpretty:
Instead of setting color and alpha on each vertex, call colorize() only once before iterating through all vertices.
In our use case, performances are multiplied by 2.


badlogic:
Good catch, thanks!

PR #653, opened 2016-07-28
[cocos2d-x] Add missing switch case

mogemimi:
This pull request adds missing switch case `SP_ATLAS_UNKNOWN_FILTER` to prevent the following warning on Xcode 7.3.1 (Clang).

```
spine-cocos2dx.cpp:42:10: Enumeration value 'SP_ATLAS_UNKNOWN_FILTER' not handled in switch [-Wswitch]
```


badlogic:
Thanks!

PR #659, opened 2016-08-02
fix: applies PR #650 correctly

ricardoquesada:
instead of passing and `int` to `takeData()` it passes a `ssize_t`
preventing a possible stackoverflow certain 64-bit platforms

related to issue #650


halx99:
It looks good to me.

PR #670, opened 2016-08-11
Add spine-ts runtime

NathanSweet:


spine-ts/webgl/src/Matrix4.ts
```diff
@@ -0,0 +1,305 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES, BUSINESS INTERRUPTION, OR LOSS OF
+ * USE, DATA, OR PROFITS) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *****************************************************************************/
+
+module spine.webgl {
+	export let M00 = 0;
```
NathanSweet: Should these guys have a type? const?

spine-ts/core/src/SkeletonBounds.ts
```diff
@@ -0,0 +1,186 @@
...
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES, BUSINESS INTERRUPTION, OR LOSS OF
+ * USE, DATA, OR PROFITS) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *****************************************************************************/
+
+module spine {
+	export class SkeletonBounds {
+		minX = 0; minY = 0; maxX = 0; maxY = 0;
```
NathanSweet: I have a feeling these should be `minx: number;` etc? Do next couple lines need types? Can't find anywhere that TS infers field type from assignment in the declaration.

spine-ts/webgl/src/Matrix4.ts
badlogic: Made const, type is inferred from initializer.

spine-ts/core/src/SkeletonBounds.ts
badlogic: TS infers field type from initializer.

http://sht.tl/udWkU0

https://www.typescriptlang.org/docs/handbook/type-inference.html

PR #672, opened 2016-08-13
[unity] Adding followBoneScale to BoneFollower.cs

Xelnath:
Used the bone's world scale value. Set to default false to prevent accidental, unwanted scale changes in active users.

Just realized this conflicts with the value of flip... adjusting the pull request.


Xelnath:
Okay, cool, I hope this works!

pharan:
I don't think this is a good addition as it is.
Following world scale rarely yields results you expect. This is more or less the same reason why Unity itself doesn't provide `Transform.worldScale`.

Following the bone local scale may be sensible though.

PR #676, opened 2016-08-19
Added canvas dir to exclude

moberghammer:


badlogic:
Ah, damn it. Good catch, thanks!

Issue #664, opened 2016-08-09
[runtimes] Root bone scaling affects animation timing


badlogic:
See http://esotericsoftware.com/forum/Scaling-Bug-6820?p=31500#p31500

badlogic:
Can kind of reproduce it with the vine sample in sfml. If i set `root->scaleX/scaleY` != 1, then the animation appears to start later.

badlogic:
Can reproduce with reference implementation. Take `vine` sample from runtimes repo, plug it into this (adapted from `SimpleTest3.java`).

```
/******************************************************************************
 * Spine Runtimes Software License
 * Version 2.3
 * 
 * Copyright (c) 2013-2015, Esoteric Software
 * All rights reserved.
 * 
 * You are granted a perpetual, non-exclusive, non-sublicensable and
 * non-transferable license to use, install, execute and perform the Spine
 * Runtimes Software (the "Software") and derivative works solely for personal
 * or internal use. Without the written permission of Esoteric Software (see
 * Section 2 of the Spine Software License Agreement), you may not (a) modify,
 * translate, adapt or otherwise create derivative works, improvements of the
 * Software or develop new applications using the Software or (b) remove,
 * delete, alter or obscure any trademarks or any copyright, trademark, patent
 * or other intellectual property or proprietary rights notices on or in the
 * Software, including any copy thereof. Redistributions in binary or source
 * form must include this license and terms.
 * 
 * THIS SOFTWARE IS PROVIDED BY ESOTERIC SOFTWARE "AS IS" AND ANY EXPRESS OR
 * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
 * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
 * EVENT SHALL ESOTERIC SOFTWARE BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
 * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
 * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
[...]

PR #680, opened 2016-08-22
[c] Implement binary skeleton loader.

Hinidu:
Hi!
I've implemented binary skeleton loader for C runtime. I think it could be useful for your users (if they choose C perhaps it is because they want to maximize performance :wink:).
Some notes:
- I've added `kvec.h` from [klib](https://github.com/attractivechaos/klib) to simplify working with dynamic arrays.
- I could avoid some allocations (mostly copying strings for names in constructors) but I chose to avoid modifying a lot of existing code to simplify this PR for now (though I would be glad to do it later).
- I think there is a problem with `vine` sample binary export - imagesPath is `./images` and mesh path is `vine` but in json export imagesPath is empty and mesh path is `images/vine`. So sfml-example crashes now on loading this attachment (other samples are rendered the same way as for json loader).


badlogic:
This looks great! I'll have to dig a bit deeper into the implementation. The use of kvec is something I'd like to eliminate if possible, though MIT and single-header file may be OK.

Thanks!

badlogic:
Another thing: we'd need you to sign the [CLA](http://esotericsoftware.com/files/cla.txt) so we can distribute your code in our runtime. You can simply fill it out digitally and send us an e-mail.

Hinidu:
> This looks great! I'll have to dig a bit deeper into the implementation.

Great! If you have any suggestions I would be glad to use them to improve this PR.

> The use of kvec is something I'd like to eliminate if possible, though MIT and single-header file may be OK.

I've used it for `readAnimation` because we don't know how many timelines it will have and for `readVertices` because we don't know how many bones and weights we will have.

Hinidu:
> Another thing: we'd need you to sign the CLA so we can distribute your code in our runtime. You can simply fill it out digitally and send us an e-mail.

Ok, but what about the sign?

badlogic:
You could just print it out, sign it, then take a photo and send us that.

Hinidu:
> You could just print it out, sign it, then take a photo and send us that.

Ok, I will try to do it today. What e-mail address I should send to?

badlogic:
contact at esotericsoftware dot com. Sorry for the trouble :/

spine-c/include/spine/Animation.h
```diff
@@ -60,6 +60,8 @@ typedef struct spAnimation {
 } spAnimation;
 
 spAnimation* spAnimation_create (const char* name, int timelinesCount);
+spAnimation* spAnimation_createWithTimelines (const char* name, float duration,
```
badlogic: Can we do without this? We'd like to keep the public API as minimal as possible.

Hinidu: We can. The only downside is the allocating of a new array for timelines instead of using existing one. Though sometimes it can be allocated with the current implementation too (if trimming of `kvec` will not be able to do it in place).

badlogic:
OK, gave a more in-depth treatment. It's awesome! Couldn't detect any obvious leaks, formatting is perfect, etc. Added a single comment regarding public API, maybe you could fix that.

Regarding kvec: we'll merge as is, and I'll add an issue for me to replace it with our own version of that functionality.

Thanks!

spine-c/include/spine/Animation.h
badlogic: Then I'd say keep the function in SkeletonBinary as a special case to increase perf, but don't expose it in Animation.h.

Hinidu: But we can trick a bit: we can call `spAnimation_create` with `timelinesCount` equals to 0 and set `timelines` and `timelinesCount` afterwards :smile:

NathanSweet:
Thanks Hinidu! You rock! :smile:

Hinidu:
> Regarding kvec: we'll merge as is

There is already many places where dynamic arrays are used in runtime. Perhaps it would be clearer to replace them with the common functionality too.

> I'll add an issue for me to replace it with our own version of that functionality.

Ahh, I see - It is not an option to ask the author of `klib` to sign your CLA...

> Added a single comment regarding public API, maybe you could fix that.

I will try to do that in the evening.

Thank you for review!

Hinidu:
> I think there is a problem with vine sample binary export - imagesPath is ./images and mesh path is vine but in json export imagesPath is empty and mesh path is images/vine. So sfml-example crashes now on loading this attachment (other samples are rendered the same way as for json loader).

And what about the `vine` binary export? Is my guess right or is there another bug?

badlogic:
I'll take a look at the vine export, it's most likely broken :)

NathanSweet:
Vine project was fixed (in `/examples`) 2 weeks ago.
https://github.com/EsotericSoftware/spine-runtimes/issues/664#issuecomment-238665550

Hinidu:
> Vine project was fixed (in /examples) 2 weeks ago.

I've got it from there and it doesn't work for me...

NathanSweet:
Which part doesn't work? I see the path as `./images/` in the [latest vine.spine](https://github.com/EsotericSoftware/spine-runtimes/raw/master/examples/vine/vine.spine) and the [atlas](https://github.com/EsotericSoftware/spine-runtimes/blob/master/examples/vine/export/vine.atlas) shows the region name as `vine` and not `images/vine`.

spine-c/include/spine/Animation.h
Hinidu: Fixed.

Hinidu:
> Which part doesn't work? I see the path as ./images/ in the latest vine.spine and the atlas shows the region name as vine and not images/vine.

@NathanSweet  Oh sorry, I forgot that I copied `vine` export before rebase on top of the master (I did checkout more than two weeks ago). After updating it from the current master everything works as it should. Thanks!

@badlogic your suggestion about `spAnimation_createWithTimelines` is implemented.

Hinidu:
I've just sent you the signed CLA :wink:

badlogic:
Thanks man, really appreciate it!

Hinidu:
> Thanks man, really appreciate it!

Thank you too! I love open source but I rarely have enough self-motivation to contribute :disappointed: The fact that we need this feature on my work has helped me to motivate myself :wink:

spine-c/src/spine/SkeletonBinary.c
```diff
@@ -0,0 +1,1009 @@
...
+			if (nonessential) {
+				mesh->edges = (int*)_readShortArray(input, &mesh->edgesCount);
+				mesh->width = readFloat(input) * self->scale;
+				mesh->height = readFloat(input) * self->scale;
+			} else {
+				mesh->edges = 0;
+				mesh->width = 0;
+				mesh->height = 0;
+			}
+			return attachment;
```
finebe: Thanks!!
Mesh type has bug. 
It is need to add a function
`spAttachmentLoader_configureAttachment(self->attachmentLoader, attachment);`

Hinidu: Gold catch! Thanks! It seems like there is already a PR that fixes this problem https://github.com/EsotericSoftware/spine-runtimes/pull/688
Btw we're still on the way to actually use it in our cocos2d game, does it work for you with this fix without other problems?

PR #688, opened 2016-08-30
fix bug when using SkeletonBinary.

boruis:
will cause crash when display skeleton using mesh.
bug


badlogic:
Thanks! I've manually merged this.

PR #694, opened 2016-09-05
[csharp] Avoid file locking in SkeletonBinary.

Hinidu:
Hello! After switching from json to binary format I found that C# runtime locks the binary file while reading it though runtime doesn't lock json files. The problem is that `SkeletonJson` uses `StreamReader` which correctly sets `FileAccess` argument but `SkeletonBinary` use `FileStream` directly and doesn't set this argument that is identical to setting `FileAccess.ReadWrite`.
Also I found some small improvements which can be achieved by specifying additional arguments for `FileStream` ctor:
- Remove `BufferedStream` because `FileStream` is buffered itself with 4096 `bufferSize` by default.
- Specify `FileOptions.SequentialScan` to give OS the chance to cache the file optimally.

### DIFF
--- a/spine-csharp/src/SkeletonBinary.cs
+++ b/spine-csharp/src/SkeletonBinary.cs
@@ -94,7 +94,7 @@ namespace Spine {
 		#if WINDOWS_PHONE
 			using (var input = new BufferedStream(Microsoft.Xna.Framework.TitleContainer.OpenStream(path))) {
 		#else
-			using (var input = new BufferedStream(new FileStream(path, FileMode.Open))) {
+			using (var input = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, FileOptions.SequentialScan)) {
 		#endif // WINDOWS_PHONE
 				SkeletonData skeletonData = ReadSkeletonData(input);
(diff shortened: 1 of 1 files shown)


pharan:
Does this have any side effects we should note?

goigle:
Wouldn't it be better to keep it consistent and update `SkeletonJson` as well?
Additionally, you only changed one `BufferedStream` to `FileStream`

Hinidu:
> Does this have any side effects we should note?

I don't aware about any side effects except mentioned above. Though I have tested this change only in our custom `SharpDX`-runtime based on `spine-csharp`. So it should be tested with `Unity` etc.

> Wouldn't it be better to keep it consistent and update SkeletonJson as well?

`SkeletonJson` uses `StreamReader` which creates `FileStream` inside the same way as I did in `SkeletonBinary`.

> Additionally, you only changed one BufferedStream to FileStream

Another `BufferedReader` wraps stream created by `XNA` framework and can not be replaced by `FileStream`.

Hinidu:
Our real issue was fixed in https://github.com/EsotericSoftware/spine-runtimes/commit/1aa8f56f9b0a180ecc24ae3c46ce5812663d8f31 so I close this PR. Thanks!

PR #697, opened 2016-09-11
Fix memory leak of c++ functional objects

halx99:
The c function ```free``` could not destruct the c++ member objects.


halx99:
@badlogic  please take a look about this bug.

PR #702, opened 2016-09-20
Fix mismatched memory deallocator in SkeletonRenderer.cpp

Rubonnek:
Currently valgrind reports the following error:

    ==3357== Mismatched free() / delete / delete []
    ==3357==    at 0x4C2B1C6: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    ==3357==    by 0x199EDB9: spine::SkeletonRenderer::~SkeletonRenderer() (SkeletonRenderer.cpp:101)

The memory which `_worldVertices` points to gets allocated in [SkeletonRenderer.cpp#L64](https://github.com/EsotericSoftware/spine-runtimes/blob/master/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L64), and therefore a `delete []` must be used instead. Otherwise we might get undefined behavior on some devices.


badlogic:
Great find, thanks!

Issue #707, opened 2016-09-23
[cocos2d-x] Animation doesn't work anymore


badlogic:
See http://esotericsoftware.com/forum/Spine-animation-stopped-showing-up-urgent-7075?p=32547#p32547

Fraggle:
I also have a very strange issue with the 3.4 runtime.
Some animation seems to show "old" triangles from another skeleton (like if the batch was not correctly resetted).

Fraggle:
I have the feeling that it was caused by this commit: https://github.com/EsotericSoftware/spine-runtimes/commit/48c5b06676c10454e52eed53dbfff7d2d6a3ee5f#diff-052b0e6f0b0ca71f0cd92b6c2e78be25

Fraggle:
I can confirm that if I revert to that commit e3cdea94a092f6ca77f36ed94832a26da5c8d6a7 (the one jut before the use of cocos2d auto-batch) then everything works ok.

badlogic:
Thanks for the info. Do you have a repro for me? The existing samples
worked as intended.

Fraggle:
Unfortunately no, i can reproduce it easily 100% of the time in our game but I cannot make a simple test case.

badlogic:
Hrm, the change in that commit is the "official" way of doing batching nowadays. Without a repro it's going to be hard to fix it. Can you describe your setup that causes the issue to appear? Maybe I can repro it like that.

Fraggle:
I think it has to do with rendering in render texture. We have a "filter" thing that will render our spine in rendertexture to apply an outline on them.
I'll try to investigate some more today and let you.

Fraggle:
It's not rendertexture.. still looking into it.
It seems to happen if I have 2 of the same spine animation in a row (that are batched together).

Fraggle:
We found the issue in our case, i'll try to explain, i think it's 100% reproductible with HEAD.

We create a spSkeletonData that is shared between multiple animation on screen (initWithData(skeletonData, false)). The skeleton data points to a spAtlas object but each animation has his own animationState.
Before the switch to use cocos batching, you used to manage your own buffers in SkeletonBatch and copy the vertices datas from the TriangleCommands.
[...]

PR #716, opened 2016-09-29
Adds Binary support to cocos2d-x

ricardoquesada:
and it also...:
- it simplifies the constructors/initializors
- the binary API is ::createWithBinaryFile()
- the JSON API is :: createWithJsonFile()
- the old API ::createWithDataFile() is deprecated and calls
createWithJsonFile()


badlogic:
Thanks! Would you be able to provide input on this issue? We've switched over to using cocos2d-x built-in batching, but are still running into issues. https://github.com/EsotericSoftware/spine-runtimes/issues/707

badlogic:
Btw, something must have gone wrong when sending the PR, it's all kinds of broken (SkeletonRenderer#initialize not being virtual, initWithJsonFile/BinaryFile not existing, etc.). Fixing it up atm, will be in master in a bit.

PR #733, opened 2016-10-18
[spine-c] fix strict C compiler error

jpoag:
Android C-Compiler build kicks this out as an error.


jpoag:
Fixed with Commit 69889d8a07d98ff910cde208c058c766a75edd1a

PR #743, opened 2016-10-23
[spine-unity] Added unscaled time functionality for canvas based skeleton

igrir:
I often used the canvas when pause ( `Time.timeScale = 0` ). Because SkeletonGraphic was using `Time.deltaTime` to update so it won't run. I changed it a little bit with `Time.realtimeSinceStartup` so the animation can run when timescale is 0. I also made it optional.


pharan:
The Unscaled time part is a good idea. I can't merge this PR because of the other conflicts though. But that feature will be in 3.6.

Note that you can also set `freeze` to `true` to prevent the SkeletonGraphic from updating.

igrir:
Whoops, I didn't know about `Time.unscaledDeltaTime`. I guess you can simply use that. Thanks 👍

PR #749, opened 2016-11-01
[spine-c] Allocate more memory for events to prevent heap corruption

Alireza-Ashayer:
When using events with large string data, the allocated memory for event is too small to store the event data. It will overflow and cause heap corruption.
It has solved our random crash problem in a game with millions of users.


badlogic:
Thanks! We are going to rework this eventually to be dynamic.

PR #759, opened 2016-11-12
[spine-unity]fix "add skeleton utility" button disappear under SkeletonAnimator

sjzlei1989:
fix: "Add Skeleton Utility" button disappear when instantiate(mecanim)
When instantiate a SkeletonDataAsset file to mecanim animation system, the button "Add Skeleton Utility" should show under SkeletonAnimator component is gone.


pharan:
This was fixed when we introduced multi-editing.
It currently casts to (Component).

PR #778, opened 2016-11-28
Skeleton json parse error

ggggamer:
Skeleton json parse error, didn't add any mix timeline.


badlogic:
Thanks, I'll apply the parsing error fix manually. However, we won't apply the compatibility fix I'm afraid.

ggggamer:
OK  I will remove this branch.

PR #788, opened 2016-12-01
Fix a bug caused a crash on VS2013 or later when in Debug mode if there is any Unicode chars such as Chinese in Atlas file.

taigacon:
In brief, the 'isspace' function allows a *unsigned* value between 0 and 255, while char is signed by default, so a char code over 0x80 represents a negative value which caused a assertion in the 'isspace' function.
The workaround is force converting the char into unsigned char.


badlogic:
Thanks!

Issue #711, opened 2016-09-27
[ts][js] Publish artifacts to NPM


scottmcdonnell:
I can't currently install spine-runtimes as a project dependency using npm - can a package.json be added?

badlogic:
We currently have no plans for this, but might do it in the future.

badlogic:
Looks like you guys resolved this without us adding a `package.json`. Good stuff! https://github.com/pixijs/pixi-spine/issues/121

dcalhoun:
@badlogic I didn't follow much of the Pixi thread discussion, but I would love to see Spine's runtimes published to npm. Manually downloading them for my project feels a bit antiquated.

scottmcdonnell:
As far as I know, all that is needed is a package.json to be created at the root. Then it can be installed directly from github using `npm install --save-dev EsotericSoftware/spine-runtimes`

dcalhoun:
@scottmcdonnell from what I understand, it's not possible to [install a subdirectory of a monorepo and it's not something that npm plans to support](https://github.com/npm/npm/issues/2974#issuecomment-169451651). So, merely adding a `package.json` to the root of this project would mean a user would have to install _every_ runtime when they need a single runtime for a project. This is far from ideal, in my opinion.

@badlogic do you have any additional thoughts on this?

scottmcdonnell:
Lots of node_modules are large packages. All runtimes would be installed
but the 'main' param in package.json would point to the JavaScript index
file.

Then a bundler such as webpack would package up just the JavaScript runtime
if 'required'

PR #793, opened 2016-12-05
Add npm support

dcalhoun:
Add package.json and npmignore to support installation of runtimes via
npm (address #711).


badlogic:
There's no way to put this in the `spine-ts` folder, correct?

dcalhoun:
### TL;DR

Not without publishing the source to [npmjs.com](https://www.npmjs.com). As referenced I referenced in #711, [npm does not support `package.json` files in the subdirectory of git repos and has no plans to change that stance](https://github.com/EsotericSoftware/spine-runtimes/issues/711#issuecomment-264238429).

### Options

1. Place a `package.json` in the root of this project, allowing users to install the runtimes with npm via this GitHub repository URL. You have the option to publish or not publish to npm. This PR accomplishes this.
1. Place `package.json` in the single directory you'd like to publish to npm (e.g. – `spine-ts`) and `npm publish` that directory. Users would be able to install from npm, but [not from GitHub](https://github.com/EsotericSoftware/spine-runtimes/issues/711#issuecomment-264238429). You would need to bump the semver of the package and `npm publish` every time you update the source for a new release.
1. Rather than publishing only `spine-ts`, you could publish all the packages to npm. You could setup [Lerna](https://lernajs.io) to publish each of the spine-runtimes to npm individually, while retaining single repository to manage the source and all issues and pull requests.

badlogic:
We've talked about this internally. We propose the following "solution".

1. Create `spine-ts/npm`, move the two files in this PR there
2. Updated `spine-ts/README.md` with a section on usage via NPM. Instruct people to copy the two files in `spine-ts/npm` to root of their clone

We'd rather not publish the spine runtimes to NPM for a) licensing reasons and b) because we'd have to start versioning the runtimes, which we currently would rather not.

Does that work for you? Would you amend your PR accordingly or should we do it?

dcalhoun:
That doesn't really accomplishing adding npm "support," but more so educating someone on how to fork a repository and add support themselves. I also find the stance against versioning odd, given that lack of versioning makes [using and following changes to the runtimes quite difficult](http://semver.org/#introduction).

I find this form of npm "support" disappointing, as it means I, and any other follower, will have to maintain a fork of this repository pulling in the latest upstream occasionally. That said, this is all your prerogative. I appreciate the time you have invested into the runtimes. I can update the PR if you desire that format.

badlogic:
I understand the frustration with versioning. The problem is that the editor version isn't fully in synch with the version of the runtime. To give a few examples:

1. We release editor version 3.5.42, but nothing has changed in master. Should we tag master with 3.5.42 and release to NPM and other artifact repositories?
2. The last editor version is 3.5.42. In the meantime, we add APIs to the runtimes, or fix bugs. This would either mean a minor or a patch version update. Should we publish runtime version 3.5.43 (patch) or 3.6.43 (API addition/breakage)?

Both the editor and the runtimes are updated at a very high frequency. We can't keep their respective versions in lock step. Would the user experience be better if the published versioned artifacts, with version numbers that don't line up with the editor versions?

If you have suggestions on how to deal with this issue, please let us know. We are open for feedback.

As for putting the NPM files in the root dir, I let @NathanSweet give some input on that.

NathanSweet:
In my experience, semantic versioning doesn't work well for all projects. Often versions end up getting bumped for breakage that affects very few users, and updating a dependency is never something to be done lightly anyway.

Mario mentioned some of the difficulties with sync'ing the editor and runtime versions. Considering that, it seems the remaining options are 1) put NPM files in a spine-ts directory and users setup NPM support themselves by moving the files, or 2) put the NPM files in the root directory. Does #2 involve keeping the version in the NPM file updated? If so, we'll need to make sure we update it when we tag the repo for a particular editor version. Littering our repo with build system files is suboptimal in general, but especially so when only spine-ts makes use of them. I suppose we've already done it for cmake, so there is precedence.

dcalhoun:
@badlogic I definitely understand the pain with attempting to keep releases in lock step. Matching versions is beneficial to users, as they can be confident they are using the correct software and runtime together. However, the work to keep things in lock step is incredibly tedious for the developers.

My initial reaction is that it is OK for the versions to not match. Even if there is _some_ confusion regarding which runtime matches a particular software, it's better than no versioning at all, in my opinion.

One idea is that you always match the [major version](http://semver.org) of software and runtimes. For example, both would be currently sitting at `3.x.x`. In semver, major versions represent backwards incompatible changes. If you introducing a backwards incompatible change, it's likely that both the software and runtime should be have their major release incremented. Outside of major versions, the minor and patch versions could differ between the software and runtime.

---

@NathanSweet I am not sure what you mean by...

> Often versions end up getting bumped for breakage that affects very few users [...]

If a version is bumped for a change (as it should to make users aware), it's still the user's prerogative if they upgrade or not. If they want a patch version bug fix, they upgrade. If they don't want to deal with a major version breaking change, they wait to upgrade.

> [...] updating a dependency is never something to be done lightly anyway.

Agreed, but the whole idea behind semver is that you can inform the user of the types of changes being made. Hypothetically, minor and patch version changes _can_ be taken "lightly," if the package developer strictly adheres to semver. I'm not naive enough to believe that it's absolute truth that there will never be an issue, but that is the spirit of semver.

> Does 2 involve keeping the version in the NPM file updated?

If you place the `package.json` in the git repo root and do _not_  publish to npm, it is _not_ required to keep the `package.json` version number up-to-date. Merely having `package.json` in the git repo will allow users to install the source from GitHub. Users can even target specific branches and commit shas.

NathanSweet:
Semver causes major version bumps from breaking changes, even for esoteric features which affect few users or for minor breakage that is very easily remedied. I prefer version numbers that communicate the general magnitude of the differences between versions. Users need to be aware of differences when they upgrade beyond what a version number can tell them. I don't find semver helpful in practice.

It sounds like #2 is the best we can do.

badlogic:
Option #2 it is. I'll try to keep the version in the package.json as up-to-date as possible. Since we don't use NPM, it'd be nice if you could help us maintain these files @dcalhoun

PR #796, opened 2016-12-06
[c] Fixed memory leak in _spSkeletonBinary_readAnimation

kikuuuty:
Releases the memory allocated when spAnimation_create is called.
```
// Animation.c(40)
self->timelines = MALLOC(spTimeline*, timelinesCount);
```


badlogic:
Thanks, I've applied this manually.

PR #800, opened 2016-12-08
[ts] Added missing return types in Utils

olssonfredrik:


badlogic:
TypeScript's type inference infers the correct types for these functions/methods.

PR #801, opened 2016-12-10
[as3] rename applyFFD -> applyDeform

bendmorris:
It seems like this MeshAttachment method should be overriding the one in VertexAttachment for linked meshes to work correctly.


badlogic:
Thanks! I've fixed this manually.

PR #813, opened 2017-01-08
Android NDK r10d  wouldn't compile, blamed wrong types

caiocsabino:
Would be cool to have those type casts added, my guess is those can also happen on other compilers.

- Added type cast to spEventType from AnimationState objects
- Added type cast to void-type realloc calls in Skeleton.c


badlogic:
Hrm, I can see why the cast is needed for the realloc calls, but for spEventType? In any case, merged, thanks!

caiocsabino:
Much appreciated!

PR #825, opened 2017-02-05
[csharp] use self stream instead of using stream for performance of u…

yy-nm:
for speed up generating spine animator in unity especially under mono runtime


yy-nm:
this only test in android and ios platform under unity

badlogic:
Thank you for your PR. I'm afraid there are multiple issues that make it unfit for a merge:

* It does  apply our coding style
* The diff for some files is a complete delete/replace making it hard to follow changes
* If we merged this, then we'd not keep the old and new way of reading from a stream, but only have the new way
* The PR makes spine-csharp unsafe just to convert the content of an int pointer to a float (for which there are safe and fast intrinsics in .NET/Mono which we use in the original code.
* It is questionable how this is supposed to improve performance. The new buffered stream introduces a level of indirection on top of the already fully in-memory MemoryStream used. The SkeletonBinaryStream use in SkeletonBinary introduces one additional level virtual calls compared to the old, static SkeletonBinary.ReadXXX, which is actually slow in Mono

Did you do any profiling on the old codeand if so, what was rhe hotspot you were trying to improve? I'm certain we can improve performance using the current code, e.g. if the int/float conversion is the culprit.

yy-nm:
In my Cause, the single spine file contains about 40 animations, 30+ slot. when i load the spine file will freeze obvious. when i use unity profile, the SkeletonBinary.ReadFloat function will be Called above 200k+, aka. Stream.ReadByte about 800K+, that occupy  1/4 cost time.

so i use custom stream class to reduce Calling ReadByte. And i just use one virtual call(aka. ReadFloat) instead of four virtual call(aka. Stream.ReadByte).

* As unsafe code about using pointer, i forget to take profiler.

* Sorry about forget to apply your coding style. this is my first PR to others.

________________________________

PR #833, opened 2017-02-16
[as3] animation totally disappears if has alpha < 1

id-ilych:
This problem was probably introduced while moving to new starling version where premultiplied alpha is no longer passed as an argument to `render()`. 

BTW if I get it right then there's no need to manually restore `blendMode` because `DisplayObjectContainer` wraps child's `render()` with `painter.pushState()`/`painter.popState()`.


badlogic:
Thanks!

PR #835, opened 2017-02-17
Should reset blend mode of batch after drawed

AyoCrazy:


badlogic:
It is assumed that the blend mode is set before drawing, by whoever uses the batcher.

PR #836, opened 2017-02-17
[ue4] Made compiling in VS work; defined behavior for GetTrackEntry()

shiverbrock:
Without this, line 52 (now line 53) of SpineSkeletonComponent.h would throw an error not recognizing USpineSkeletonDataAsset.

Also, attempting to use GetTrackEntry() from SpineSkeletonAnimationComponent.h would throw a Linker error as the behavior was not defined.


badlogic:
Thanks, this is already fixed in the 3.6-beta branch.

PR #838, opened 2017-02-21
[ts][Don't Merge] A bug of SkeletonJson.

natural-law:
I found that there is a bug of SkeletonJson:
Once the parameter of `SkeletonJson.readSkeletonData` is an cached object & the Skeleton Data has mesh attachment & the `SkeletonJson.scale` is not 1 . The cached json object will be modified by the `SkeletonJson.readSkeletonData`.

If the cached json object is used once more, the mesh attachments will be shown with wrong effect.

This PR modified the test.html which can reproduce the problem. Sorry for my poor English! Looking forward your reply!

### DIFF
--- a/spine-ts/webgl/example/test.html
+++ b/spine-ts/webgl/example/test.html
@@ -14,6 +14,7 @@
 var canvas, gl, renderer, input, assetManager;
 var skeleton, bounds, state;
 var timeKeeper;
+var skelJsonObj;
 
 function init() {
 	canvas = document.getElementById("canvas");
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks for the reproduction case. I've fixed this in the [3.6-beta branch](https://github.com/EsotericSoftware/spine-runtimes/commit/0b8913691a4ffc4443b9adda167b62f53c9c369b), it should be easy to port over to the master branch in your fork until we have released 3.6.

natural-law:
@badlogic Thanks for your quickly reply.

PR #848, opened 2017-03-01
check null

dorisugita:
coverity doesn't good at this.


badlogic:
Coverity is wrong in this case, kvp.Value can not be null.

PR #849, opened 2017-03-01
check null

dorisugita:
if region is null, it might excepts errors.


badlogic:
Again, a false positive of Coverity.

PR #871, opened 2017-04-06
Remove dipping in interruped mix

pharan:
If there's no multipleMixing == false, the mix progress should not be inherited as mixAlpha since this will cause a dip in the common cases where a mix is interrupted.


NathanSweet:
Merged manually to rearrange some logic, thanks!

PR #873, opened 2017-04-08
[lua][corona] Performance optimizations

torbenratzlaff:
# Several major and minor performance and memory optimizations.
Although github shows that 523 files have been changed this is **not** the case, as it's the result of a not so well resolved merge conflict (and I was unable to resolve that issue). Only files in the lua and corona runtime are affected.

The changes include:
- localization of table fields to prevent multiple field indexing in time critical processes
- replacement of most ipairs-iterator with simple for-loops
- implementation of metatable to establish an oop approach for objects that are created in large numbers (like bones, animations etc.)

All changes where tested with the provided examples and several other self created test situations. As the runtime is so extensive further testing is necessary.


badlogic:
Could you rebase your branch on the latest master branch? Alternatively, could you send a plain patch file between current files in master and your tree changes?

badlogic:
Nevermind, I see the individual commits. The changes in formatting make it hard to follow the actual changes.

Do you have any benchmarking data that demonstrates the performance improvements? This is a big change to the Lua based runtime(s). Modifying this dynamically typed codebase without tests that execute all codepaths and validate behaviour is a bit scary, especially if the gains aren't there.

Some if the ipairs iterations actually rely on the behaviour of ipairs regarding "holes". It was also my understanding that ipairs aren't slower than standard iteration according to the Lua performance guide (been a while, I might remember this wrong).

badlogic:
I went through the changes, thanks for this work! We did some major changes in the beta branch. I'll manually move your master changes over and will do some benchmarking on my end.

torbenratzlaff:
Hey badlogic,
thanks for reviewing my changes (especially as they are so fucked up) :)
I'm a litle bit confused now, should I still do any of the things you wrote in your frist two comments? (or is it obsolete as you did or will do it yourself?)

A simple for-loop is a little bit faster then ipairs (but not that much, take a look at: http://stackoverflow.com/questions/8955085/should-i-use-ipairs-or-a-for-loop ). 
I didn't know there could be holes in the data, as it seemed to me there are no holes in arrays from the spine export. Or do you use ipairs because the data could be changed manually and therefore get holes during runtime?

badlogic:
No need for you to do anything!

Yes, I use ipairs due to the possibility of holes being introduced if users
do low level stuff. Shouldn't happen, but can.

badlogic:
I've applied some (but not all) of the enhancements in your commits. The for loop changes didn't make any noticable difference in my tests (tested with the Lua VM that comes with Corona and Love2D). I've also refactored the rendering code for Corona, which should now be quite a bit faster! That had the biggest impact regarding performance.

Thanks for your PR!

Issue #675, opened 2016-08-16
[ts] Handle context loss in WebGL backend and widget


badlogic:
See http://esotericsoftware.com/forum/ts-webgl-Handling-Context-Lost-6861

badlogic:
Users can theoretically re-create all resources themselves on context loss. However, it'd be nice to have a simpler mechanism so you don't need to recreate object instances. Problem is that all WebGL "classes" require a handle to the WebGLRenderingContext and must recreate meshes/textures/shaders as they were constructed (e.g. with/without mipmapping). Which means those "classes" need to store more info to recreate things.

geddski:
@badlogic I'm hitting this in my app now after navigating a few times (single-page app). What do I need to handle/ recreate in order to avoid the app blowing up?

badlogic:
Fixed in the 3.6-beta branch. See commit 2b4c2bcf94def732caa31420aca4a6d7ed1ac6a2 and [CHANGELOG](https://github.com/EsotericSoftware/spine-runtimes/blob/3.6-beta/CHANGELOG.md#webgl-backend)

geddski:
Sweet I'll try it out now. Thanks @badlogic

geddski:
Sweet I'll try it out now. Thanks @badlogic

geddski:
@badlogic this totally fixes it! You're a genius.

geddski:
@badlogic shoot, still errors out in Chrome on Windows. Chrome on Mac works fine. Edge and Firefox on Windows work fine.

geddski:
```js
Uncaught TypeError: Cannot read property 'VERTEX_SHADER' of null
```

badlogic:
When does this happen? After switching tabs a bunch of times while having a Spine widget on one of the tabs? Using 3.6-beta?

geddski:
yep using 3.6-beta, happens not when switching tabs but when going to new routes in my single page app, where old contexts are discarded and new ones are created. Same error that used to happen in all the browsers before your fix. Now it's just in chrome.

badlogic:
I've tried to recreate the issue, see this new test https://github.com/EsotericSoftware/spine-runtimes/commit/66c6d5cd899400b3770c62afb7f719c67060929d

Upon clicking the reload button, the old 5 widgets are removed from the DOM (after pausing them, so no further calls to requestAnimationFrame are made), and 5 new widgets are added to the DOM.

[...]

PR #876, opened 2017-04-13
reuse existing canvas element when possible

geddski:
When using Widget with React and other frameworks it's common to change properties but reuse the same DOM element. But when creating a new widget while reusing the same DOM element, spine-widget creates a new canvas object each time. This results in one element containing many canvases at once. 

This solution checks for an existing canvas to use instead of assuming a new one needs to be created each time. Makes Widget much more React friendly.

### DIFF
--- a/spine-ts/widget/src/Widget.ts
+++ b/spine-ts/widget/src/Widget.ts
@@ -60,10 +60,14 @@ module spine {
 
 			this.validateConfig(config);
 
-			let canvas = this.canvas = document.createElement("canvas");
+			let existingCanvas = <HTMLCanvasElement> element.children[0];
+			let canvas = this.canvas = existingCanvas || document.createElement("canvas");
 			canvas.style.width = "100%";
 			canvas.style.height = "100%";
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks! Sorry for the merge taking a bit.

badlogic:
FYI, I'm working on a different solution to retina screens in the 3.6-beta branch (alongside a fix for #675)

badlogic:
Here's the more robust fix using the device pixel ratio https://github.com/EsotericSoftware/spine-runtimes/commit/355d80129ff4b28fa0077c99017ccade5998d2b6#diff-5ce36b51693f3dbf8e395d28099ededcR248 We'll merge this into master once 3.6 is released.

geddski:
Ok cool. That's pretty similar to how I was handling retina. I'll try this and report back.

badlogic:
Your method used the bounding box of the skeleton, which is not necssarily
connected to the resolution of images used by attachments. This should be
the "proper" solution, using the density of the display device.

geddski:
@badlogic I made a new discussion for retina in #881

PR #877, opened 2017-04-13
Supporting ETC1 Alpha channel 

Kameiha:
For supporting ETC1 Alpha channel, we should pass whole "texture" to "trianglesCommand", not only its id, so that we can obtain the alpha texture of the texture in the triangles command.


halx99:
Great job.

badlogic:
I've added this manually to the 3.6-beta branch, see https://github.com/EsotericSoftware/spine-runtimes/commit/26c1bfe8b4d9daa56b4dcc6d5fc0e4a701005a5f

Thanks!

Issue #899, opened 2017-05-11
[runtimes] Animation change mixing problem (unnecessary motion between two similar poses)


snake5:
When I change an animation with mixing, the expected behavior is shortest-path interpolation between the two animations. The actual behavior is that additional motion appears between two nearly identical poses.

I am using the spine-c runtime (downloaded early this year).
There are two animation tracks used (body id=0, arm id=1), mix duration is 0.1 (but increasing it didn't help).
Character: [player_char.zip](https://github.com/EsotericSoftware/spine-runtimes/files/992262/player_char.zip)
Animations:  ladder_prepjump, ladder_jump

Recording of the behavior: ![spinebug1crop](https://cloud.githubusercontent.com/assets/1819544/25939045/25a77b36-363a-11e7-8a24-63383df1be0a.gif)
Except the beginning, gif consists of switching between the two animations. There should be almost no motion but the actual result is unwanted, immersion-breaking deformations.

I did some debugging and found out that it helps with this issue to use alpha=1 when setupPose!=0 but it also breaks a bunch of other things so I haven't yet found a good workaround.
Initially the first keyframe was the same between both animations but I tried changing it, which didn't help also.
Since it is quite difficult to find a workaround for this issue, I would suggest that this is a high-priority issue.

snake5:
Searched the forums, found this:
http://esotericsoftware.com/forum/Mix-creating-unexpected-behavior-in-new-version-7412?sid=cc7dad20b4bd159d49060d42c7dab26e&p=37952&hilit=mixing+bug#p37952
https://gist.githubusercontent.com/pharan/cffdc3b2aee9288c9ca8b4ab94ee97d2/raw/ec10be8fe92900d1dcb6bc22cad0a1e6fa21c803/AnimationState.cs

It appears to be different from spine-c and the differences suggest that the issue might be fixed there. I'll check later.

snake5:
Looks like this issue is fixed in 3.6 beta.

snake5:
Unfortunately I've found another bug of the same kind in latest 3.6-beta source.

When mixing overlaps (like when quickly switching the animation twice), there is still unwanted jittering.

https://drive.google.com/open?id=0Bz3gNbfHFo_lMU14V1FIWjY4ckU
Please download the video as the built-in player is showing a damaged copy instead of the 60fps original.

[...]

PR #900, opened 2017-05-12
[c] Fix incorrect blending of translate timelines

Jimbly:
I just updated from a rather old version of spine-c, and animation blending appeared to not be working right.  In my specific case, I have a bone at X=200, and two animations, both of which have a translation of -200, so, obviously, whatever the alpha between the two, the result should be 0.  With an alpha of 0.5, the old code would set bone->x = 200 + -200 * 0.5 (100) on the first pass, and then bone->x += (200 + -200 - 100) * 0.5 -> bone->x += -50 -> bone->x = 50!

The alpha should simply be applied to the translation coming out of the animation, the second pass should not use the results of the first pass to decide how much of the second pass to apply!  In theory, this should be commutative (A * 0.2 + B * 0.8 == B * 0.8 + A * 0.2), and this fix also ensures that.

Just glancing at it, I *think* _spScaleTimeline_apply has a similar issue (if bone->data->scale is 100, and two animations are each scaling to 1/100, it's going to blend to a scale of, maybe 25.75?), but it's more complicated with the mixingOut stuff and I don't have a good example set up for that.

### DIFF
--- a/spine-c/spine-c/src/spine/Animation.c
+++ b/spine-c/spine-c/src/spine/Animation.c
@@ -343,12 +343,11 @@ void _spTranslateTimeline_apply (const spTimeline* timeline, spSkeleton* skeleto
 		y += (frames[frame + TRANSLATE_Y] - y) * percent;
 	}
 	if (setupPose) {
-		bone->x = bone->data->x + x * alpha;
-		bone->y = bone->data->y + y * alpha;
-	} else {
-		bone->x += (bone->data->x + x - bone->x) * alpha;
-		bone->y += (bone->data->y + y - bone->y) * alpha;
+		bone->x = bone->data->x;
+		bone->y = bone->data->y;
 	}
+	bone->x += x * alpha;
+	bone->y += y * alpha;
 
 	UNUSED(lastTime);
 	UNUSED(firedEvents);


Jimbly:
Hmm, though the existing code is definitely wrong, my fix also fails in some other cases, I think when mixing out other animations (presumably when it doesn't get a "setupPose" every frame), I'll investigate further.

Jimbly:
After investigating further, I believe this fix is correct.

The other issues I ran into all stemmed from bugs in mixing logic (similar, but unrelated, issues where the sum of alphas used on a frame did not sum to 1.0).  The other fixes I'm way less confident in (and they don't meet the style of this repo), though it seems to fix a bunch of issues I was having since migrating to the Spine 3.x runtime, and those fixes are over at https://github.com/Jimbly/spine-runtimes/commit/82db19b9c30f8bbaa59dbe81237dd553b854f972 - we can open another issue for those if needed.

NathanSweet:
> In theory, this should be commutative (A * 0.2 + B * 0.8 == B * 0.8 + A * 0.2), and this fix also ensures that.
AnimationState doesn't do this kind of "additive" mixing, which has its own problems. There was some discussion about using additive [here](http://esotericsoftware.com/forum/Code-HARDCORE-the-mecha-game-6553), but the OP went MIA and nothing came of it.

AnimationState uses "tracks", where animations are applied on top of each other, enabling partial or complete overriding of the animations applied on lower tracks. I believe TranslateTimeline is correct without this PR.

Note that AnimationState in the `3.6-beta` branch has a "multiple mixing" setting:

When false, only two animations can be mixed at once. Interrupting a mix by setting a new animation will discard one of the two old animations, keeping the one closest to being fully mixed in. Discarding an animation in this way may cause keyed values to jump.

When true, any number of animations can be mixed at once without causing keyed values to jump. Mixing is done by mixing out one or more old animations while mixing in the newest one. When animations key the same value, this may cause "dipping", where the value moves toward the setup pose as the old animations mix out, then back to the keyed value as the new animation mixes in.

There is currently some activity in #899 about removing the multiple mixing setting by finding a way to have both non-dipping and multiple mixing.

Jimbly:
The kind of mixing being down currently does make sense for adding tracks on top of each other, but this code is also being used for blending/mixing animations in a single track, which seems mathematically wrong for that case.  Perhaps one kind of mixing should be done when blending animations on a single track (also would expect alpha to always sum to 1.0 in that case), and then other kind of mixing when mixing the different tracks together.

That being said, the Spine 2.x runtime I was using before appears to have been doing the same kind of low-level mixing as the 3.5 runtime, however it did *not* have this problem on identical data, somehow it didn't end up having these artifacts ("dipping"?).  Not sure why that is, though!

NathanSweet:
That's true, maybe additive could be used intra-track. We are still working on the AnimationState in 3.6-beta and would like to remove the multiple mixing setting once we solve supporting no dipping and multiple mixing.

2.x didn't support multiple mixing (doesn't dip). 3.5 supports *only* multiple mixing (dips). 3.6-beta makes non-multiple mixing the default (doesn't dip) and has an option for multiple mixing for those users that need it (dips). I suggest grabbing the AnimationState from the 3.6-beta branch, which should be a drop-in replacement and work like you want.

Jimbly:
The AnimationState form the 3.6-beta branch, plus rolling back this change, does seem to work *better*, but still is causing incorrect results when making quick transitions, maybe when more than one animation is mixing out or probably related to when the most recent animation wasn't fully mixed in when it started mixing out (the other thing I tweaked with in my [branch](https://github.com/Jimbly/spine-runtimes/commit/82db19b9c30f8bbaa59dbe81237dd553b854f972)), so I think I'll just stick with my branch for now, but will hopefully be able to use 3.6 when it's ready.

NathanSweet:
You'll get snapping (with multiple mixing off, which is default) when interrupting a mix. If you see incorrect behavior in any other cases, it would good to know about. The best way to report that is to reproduce it in the 3.6-beta [Skeleton Viewer](http://esotericsoftware.com/spine-skeleton-viewer) and provide JSON, Skeleton Viewer settings, and exact repro steps. If it happens with quick transitions, it can help to reduce the `Speed` in Skeleton Viewer.

Jimbly:
That's definitely one of the two things I fixed/worked around in my branch, the other was related - when blending from animation A to animation B, and then going to C, which does not specify any blending, some of A was still in effect.  Both the dipping and this issue were fixed in the same way (recursively applying the alpha down through _spAnimationState_applyMixingFrom calls, and also using the return value in the recursive call just like how it gets used in _spAnimationState_apply).  I've got a release planned this week, so don't have any more time to spend on this at the moment, but if I get some time, I'll try and see if the issues happen within the Skeleton Viewer (I guess would involve manually changing the mix slider between clicking animations, as the json format still doesn't allow you to specify per-animation blending information, I'm guessing).

NathanSweet:
AnimationState should never be leaving an animation partially applied. The last time an animation is applied it should set everything back to the setup pose. If it is leaving the skeleton in a bad state, we'd love to fix it. I certainly understand having limited time though!

Thinking about using additive intra-track, we'd still need to combine tracks using our current (layered) mixing. That likely means we'd need to calculate the poses separately for two tracks using additive, then combine them using layered. Juggling that much state is a bit of a pain. I'd love to just use additive, as that solves a lot of problems, but layering is very useful and additive does have a way of overwriting the pose other animations have applied.

We've done some work on AnimationState improvements, though it's still a WIP: b882fb8c51787eee27f88433252a6de16b77b134

PR #917, opened 2017-06-02
[unity] Fixed Unity 5 compiler condition to include Unity 2017 compatibility

t04glovern:
In Unity 2017 the **UNITY_2017** compiler flag can be used, however this flag is not available in versions prior. **UNITY_5_3_OR_NEWER** resolves this problem for now.


pharan:
Thanks! We're doing a pass on the compiler directives, probably next week.

pharan:
This was done. Thanks.

PR #922, opened 2017-06-11
[ue4] make the plugin build on UE 4.16.x

NuLL3rr0r:
Any project using ue4-spine plugin fails to build on UE 4.16.x unless the build system is modified according to the new changes in the UE4 build system.


badlogic:
Hrm, didn't fail in 4.16.0 for me, only printed deprecation warnings. Thanks!

badlogic:
Hrm, didn't fail in 4.16.0 for me, only printed deprecation warnings. Thanks!

NuLL3rr0r:
Thank you for the quick response and merge. I'm using 4.16.1 on Linux and the builds stops with an error after the deprecation warnings. I have quite a few settings in my .Build.cs. Maybe somethings made those deprecation warnings to be treated as errors. Thanks anyway.

PR #935, opened 2017-07-10
dead code

dorisugita:
vertices is always null.


pharan:
Looks like a misport. This isn't quite the fix but we'll fix it soon.

pharan:
Thanks for reporting!

dorisugita:
@pharan  Thank you!

PR #936, opened 2017-07-10
check null

dorisugita:
`GetString(map, "end", null)` might be null

### DIFF
--- a/spine-csharp/src/SkeletonJson.cs
+++ b/spine-csharp/src/SkeletonJson.cs
@@ -420,8 +420,10 @@ namespace Spine {
 					ClippingAttachment clip = attachmentLoader.NewClippingAttachment(skin, name);
 					if (clip == null) return null;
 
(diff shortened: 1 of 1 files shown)


badlogic:
@pharan seems like Nate changed end slot detection in reference implementation. https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/SkeletonJson.java#L414 Just port that over. I'll check the other runtimes.

badlogic:
Just checked, it's correct in all other runtimes. The PR is almost correct, but not quite. @pharan please just port over the ref impl.

badlogic (commented):
Does not quite match reference implementation https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/SkeletonJson.java#L414 @pharan to the porting machines!

PR #929, opened 2017-06-27
 Added getTexture() method

aram-ahak:
We need use our textures repository, that's why i added getTexture method in the StarlingAtlasAttachmentLoader class.

PR #952, opened 2017-07-24
Removed TextureAtlas operations in StarlingAtlasAttachmentLoader 

aram-ahak:
My pull request ( https://github.com/EsotericSoftware/spine-runtimes/pull/929) merged partialy, but it didn't help me. Please merge this pull request if it possible.

We have textures dictionary in our project,  and we get the texture from dictionary, that is why in the PR I asked you add the getTexture() method. We  not necessary to know textureAtlas of the texture. so we need that the newMeshAttachment and newRegionAttachment methods did't operate with TextureAtlas. 
Below is presented how i whant to use StarlingAtlasAttachmentLoader 
`
public class CustomStarlingAtlasAttachmentLoader extends StarlingAtlasAttachmentLoader {

    public function CustomStarlingAtlasAttachmentLoader() {
        super(null);// we don't use atlas
    }

    override protected function getTexture(path:String):Texture {
        return GPURepository.getTexture(path);
    }

}
`

`
public class Test extends Sprite {

	public function Test () {
        var attachmentLoader:AttachmentLoader = new CustomStarlingAtlasAttachmentLoader ();
        var json:SkeletonJson = new SkeletonJson(attachmentLoader);
        var skeletonData:SkeletonData = json.readSkeletonData(jsonData);
         .............................
         }
   }
`

### DIFF
--- a/spine-starling/spine-starling/src/spine/starling/StarlingAtlasAttachmentLoader.as
+++ b/spine-starling/spine-starling/src/spine/starling/StarlingAtlasAttachmentLoader.as
@@ -59,11 +59,11 @@ package spine.starling {
 		}
 
 		public function newRegionAttachment(skin : Skin, name : String, path : String) : RegionAttachment {
-			var texture : Texture = getTexture(path);
+			var texture : SubTexture = getTexture(path) as SubTexture;//atlas.getTexture() method always return type of SubTexture
 			if (texture == null)
 				throw new Error("Region not found in Starling atlas: " + path + " (region attachment: " + name + ")");
 			var attachment : RegionAttachment = new RegionAttachment(name);
-			var rotated : Boolean = atlas.getRotation(path);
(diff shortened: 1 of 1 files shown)


badlogic:
This breaks other uses of the attachment loader. I think it makes sense for you to have your own attachment loader implementation specific for your project. What do you think?

aram-ahak:
I don't think that this breaks other uses. What do you mean about own attachment loader implementation?I must duplicate methods of StarlingAtlasAttachmentLoader?

badlogic:
I've manually merged this with some minor modifications (cleanup). Thanks!

aram-ahak:
Sorry I forgot improve code on line 105. Please consider var rectRegion : Rectangle = texture.region; instead var rectRegion : Rectangle = atlas.getRegion(path);

PR #964, opened 2017-08-04
Single precision: Backport from pixi-spine

ivanpopelyshev:
The explanation is here: https://github.com/pixijs/pixi-spine/pull/177 by @paolotozzo

> This is to fix a bug I've noticed when putting an event to get triggered at the very end of an animation and when there's support for Float32 Arrays. Basically I've discovered that in that case it can happen that the condition at the line 291 of AnimationState.ts (method queueEvents)
> 
> if (event.time > animationEnd) continue;
> 
> passes.
> 
> The issue here is that all timelines objects (including the EventTimeline) have a property frames that is a Float32 Array.
> When setting the timing for each frame with setFrame the value / item of the array gets converted in its single precision float representation, which means that for example an event put in the last frame with time: 4.5333 gets converted to time: 4.533299922943115, which makes the condition above true as animationEnd is actually the time value from the very last frame, whereas event.time is the value passed in the constructor when building the whole Spine (new core.Event(eventMap.time, eventData)), and that in this case would be 4.5333
> 
> To fix this I've used the Math static method fround (see here) which is used to set the time of the event, which means only at the time of the initialisation. Basically we are making consistent the number type (or better, the number representation) so when is time to compare them, the behaviour is always the expected one.
> 
> The method is widely supported, but as IE doesn't, I've used the recommended polyfill.
> I hope it is all clear.


ivanpopelyshev:
http://docs.w3cub.com/javascript/global_objects/math/fround/

badlogic:
Thanks!

PR #966, opened 2017-08-11
fix clipping issue

zzya:
Now,I think should add "spSkeletonClipping_clipEnd" before every "continue", because I encountered such a problem in our project. There will be some slots without attachment,if skip them, the clipping will have some problems.


badlogic:
You are actually correct on this one! Good catch, thanks for sending the PR!

PR #974, opened 2017-08-28
Use Epsilon

dorisugita:
It is better.


badlogic:
Mathf is a Unity API and can't be used in the engine independent C# runtime.

pharan:
There is a [System.Single.Epsilon](https://msdn.microsoft.com/en-us/library/system.single.epsilon(v=vs.110).aspx) but I don't know the repercussions of using that in terms of runtime consistency.

PR #976, opened 2017-08-30
Adds a new API spTrackEntry_setDisposeCallback for notifying TrackEntry will be disposed.

dumganhar:
This method could be used for releasing script binding object.
Since we need to use native objects to control the life cycle of JS objects, and TrackEntry is exposed to script, spine runtime needs to notify that TrackEntry is disposed,
in that case, script object will be safe to be released in the callback of `spTrackEntry_setDisposeCallback`.

### DIFF
--- a/spine-c/spine-c/include/spine/AnimationState.h
+++ b/spine-c/spine-c/include/spine/AnimationState.h
@@ -150,6 +150,9 @@ SP_API float spTrackEntry_getAnimationTime (spTrackEntry* entry);
 /** Use this to dispose static memory before your app exits to appease your memory leak detector*/
 SP_API void spAnimationState_disposeStatics ();
 
+typedef void (*TrackEntryDisposeCallback)(spTrackEntry*);
+SP_API void spTrackEntry_setDisposeCallback(TrackEntryDisposeCallback cb);
+
 #ifdef SPINE_SHORT_NAMES
 typedef spEventType EventType;
 #define ANIMATION_START SP_ANIMATION_START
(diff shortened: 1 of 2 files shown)


badlogic:
`spTrackEntry` has a `listener` field you can set. The field is typed to a function that gets a range of events, including a track entry dispose event. Let me know if that doesn't work for your use case and why.

dumganhar:
Oh, I didn't find that method. The `listener` callback method seems good.
But I still have a question. It looks like not all dispose events of TrackEntry instances will be notified by that listener.
For instance,

spAnimationState_dispose -> _spAnimationState_disposeTrackEntries -> _spAnimationState_disposeTrackEntry -> Just free memory, no callbacks.

No dispose callbacks will be invoked, right?
I may be wrong since I am not familiar with the logic in spine runtime. Plz correct me.

dumganhar:
BTW, I'm using the version: 3.5.35: https://github.com/EsotericSoftware/spine-runtimes/releases/tag/3.5.35

badlogic:
You are correct, the dispose event will not be fired for left-over track entries in `spAnimationState_dispose`. I think it makes sense to fire it in that function, but that'd slightly change behaviour with respect to other runtimes. @NathanSweet what do you think?

NathanSweet:
I think it makes sense for all runtimes which have an explicit AnimationState `dispose` to fire `end` for the current entry for each track and `dispose` for both the current and queued entries.

badlogic:
I shall make it so then.

badlogic:
@dumganhar I've now added calls to the listeners in case the `spAnimationState` is disposed, see https://github.com/EsotericSoftware/spine-runtimes/commit/a4a97969e8916be7d6ba531707b662ebbad8da9b.

dumganhar:
@badlogic , thank you. That should be able to work. 👍

PR #980, opened 2017-08-31
Fixed SkeletonRendererInspector UI

ZimM-LostPolygon:
I'm pretty sure this was not the expected look:
default
So, fixed that.


pharan:
Thanks for the PR, @ZimM-LostPolygon . 

It's supposed to be a horizontal scope because the reload button is right beside it.

That's really weird. Why is the SkeletonDataAsset field showing the inspector for the whole SkeletonDataAsset instead of just the reference field?

What version of Unity is this? Or what other plugins do you have? I can't seem to replicate.

ZimM-LostPolygon:
Right. After investigating a bit, I've found a third-party editor script that was overriding all ScriptableObject inspector drawing. Not Spine's fault, closing this.

PR #989, opened 2017-09-13
Fix for cocos2d-x etc1 builtin alpha support.

halx99:


badlogic:
Looks good. Could you sign our [CLA](http://esotericsoftware.com/files/cla.txt), scan it or take a photo, and sent the scan to contact@esotericsoftware.com? Cheers!

halx99:
@badlogic  done.

badlogic:
Thanks! I've manually merged the PR to both the 3.6 and the 3.7-beta branch. There's a subtle thing going on in case the skeleton uses a multi-page atlas: the first texture found must be ETC1 for this to work. I think we can assume that multi-page atlases will not have mixed ETC1/non-ETC1 pages though.

halx99:
Yes, I think so.

sp-dani-garcia:
Hi,

@badlogic in the integration of this changes you forgot to add the calls to [setupGLProgramState](https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L74). This calls are missing [here](https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L71) and [here](https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L743).

Also, it's probably that the [TwoColorTintProgramState](https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L741) doesn't support ETC1 as is not changed at all in this PR.

I think that your assumption that multi-page atlases won't mix ETC1 and non-ETC1 is correct 👍 .

Thank you.

badlogic:
Fixed, thanks!

badlogic:
Two color tint indeed doesn't support ETC1+alpha at the moment.

Issue #992, opened 2017-09-14
Check proper tint black blending for non-PMA images


NathanSweet:
We should verify non-PMA blending is incorrect when using tint black. Consider a single formula to combine PMA and non-PMA cases, where `u_pma` is a uniform which is 1 for PMA and 0 for non-PMA:
```
gl_FragColor.rgb = (u_pma + (1.0 - u_pma) * texColor.a - texColor.rgb) * v_dark.rgb +
                   texColor.rgb * v_light.rgb; 
```

ivanpopelyshev:
Ok, so, which branch should i make PR against?

Whole idea of two-tint looks like that: We have to map [0..1] color into [dark..light], independent from alpha. Alpha works like always. There's no "dark alpha" nor"light alpha", just one alpha.

# Premultiplied

BlendMode : (1, 1-src_alpha)
v_light, v_dark are premultiplied by alpha, u_pma is 0.0 and is packed as a "v_dark.a" but it has no logic relation to the dark side. Don't look at that its used in same multiplication as v_dark, its just a coincindence, there's no such thing as "dark alpha"

# NON-PMA

BlendMode : (src_alpha, 1-src_alpha)
u_pma is 1.0, v_light and v_dark are non-premultiplied.

ivanpopelyshev:
Also, you guys are much better than me in testing, I'm just a math guy. If we succeed, I'll add it to pixi.js and may be I add it to CanvasRenderer too! Also we'll try bunnymark two shaders, may be we can use two-tint shader by default.

Btw, tint can be packed into 4 bytes, and dark-light into 8. Right now runtime is packing it into 4 floats, which is not optimal.

ivanpopelyshev:
Ok, now PMA is 1.0 for premultiplied. Please check it in PR.

badlogic:
I've tested the PR by @ivanpopelyshev and it works as intended for tint black in both PMA and none PMA mode. However, this issue is about verifying that non-PMA tint black is broken. @NathanSweet any pointers on how to do that?

ivanpopelyshev:
something like TEXTURE=0x707070, TEXTURE_ALPHA=0.5 with DARK=0x707070 will show the difference in both pma and non-pma modes. Current formula is just wrong for both cases.

badlogic:
I wrote a quick test https://github.com/EsotericSoftware/spine-runtimes/blob/3.6/spine-ts/webgl/example/test-pma-tintblack.html

[...]

PR #993, opened 2017-09-14
spine-ts: fix dark in both pma and non-pma

ivanpopelyshev:
#992 

According to my calculations, both modes were wrong. I do not know about other runtimes.

To make shader work in both modes, I've added v_dark.a = 1.0 in PMA, 0.0 in non-PMA.

### PMA mode

- FIX 1.0 switched to texColor.a , because that's what math says
- IMPROVEMENT premultiply dark.rgb to alpha

### NON-PMA

- FIX alpha should not affect rgb at all

### DIFF
--- a/spine-ts/webgl/src/Shader.ts
+++ b/spine-ts/webgl/src/Shader.ts
@@ -247,9 +247,8 @@ module spine.webgl {
 
 				void main () {
 					vec4 texColor = texture2D(u_texture, v_texCoords);
-					float alpha = texColor.a * v_light.a;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Its interesting that the same thing can be reproduced with http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/geom/ColorTransform.html , however, it would not allow to invert the colors. At the same time, dark-light wont allow to decrease the specific color.

badlogic:
Thank you kind sir!

PR #1014, opened 2017-10-09
Fixes heap corruption with more than 1000 mesh vertices (cocos2dx) 

daniele-rapagnani:
We are using the cocos2dx runtime for one of our games and we have been experiencing odd crashes and bugs that looked like the result of heap corruption. Since they seemed to happen after a particular animation was played I started investigating the Spine runtime and eventually found that  the world vertices array was hardcoded to 1000 vertices and there was no check in place to at least crash the application with a clear error message when the code addressed more than 1000 vertices, not even an assert. We had 1002 vertices and this caused really nasty bugs all over the place which took a lot of time to track down.

I've made an attempt to fix the issue by pre-determining the array size when the skeleton data is assigned to the SkeletonRenderer instance. The array is still reused and resized at runtime if needed by doubling its size to minimize the chance of frequent resizing. I don't know if this is the right approach with the issue because I do not have much experience with the internal workings of the Spine runtime, but we have been using this solution for a while and it seems to work just fine so I'm submitting it.

### DIFF
--- a/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
+++ b/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
@@ -61,8 +61,6 @@ SkeletonRenderer* SkeletonRenderer::createWithFile (const std::string& skeletonD
 }
 
 void SkeletonRenderer::initialize () {
-	_worldVertices = new float[1000]; // Max number of vertices per mesh.
-	
 	_clipper = spSkeletonClipping_create();
 
 	_blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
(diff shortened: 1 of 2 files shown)


badlogic:
Sorry, that was a clear oversight on our end. Your solution works, but is not 100% correct. At the time of ``setSkeletonData` not all attachments may be set on slots yet, so calculating the maximum number of world vertices required may be wrong. The dynamic resizing must actually happen in the rendering method.

I'll manually fix this up. I've created an issue for this here https://github.com/EsotericSoftware/spine-runtimes/issues/1015

Thanks for bringing attention to this!

PR #1018, opened 2017-10-12
[cocos2dx] allow load custom atlas texture in _spAtlasPage_createTexture

jareguo:
This PR is for CocosCreator. In CocosCreator 1.8, atlas texture will be preloaded by scene, so it will not cached in TextureCache. We want to add a custom loader to get preloaded textures from Creator. Thanks!

### DIFF
--- a/spine-cocos2dx/src/spine/spine-cocos2dx.cpp
+++ b/spine-cocos2dx/src/spine/spine-cocos2dx.cpp
@@ -31,6 +31,13 @@
 #include <spine/spine-cocos2dx.h>
 #include <spine/extension.h>
 
+namespace spine {
+	static CustomTextureLoader _customTextureLoader = nullptr;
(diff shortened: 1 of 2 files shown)


spine-cocos2dx/src/spine/spine-cocos2dx.cpp
```diff
@@ -31,6 +31,13 @@
 #include <spine/spine-cocos2dx.h>
 #include <spine/extension.h>
 
+namespace spine {
+	static CustomTextureLoader _customTextureLoader = nullptr;
+	void spAtlasPage_setCustomTextureLoader (CustomTextureLoader texLoader) {
+		_customTextureLoader = texLoader;
```
jareguo, line 37: In CocosCreator, this method will be called with a custom load function.

badlogic:
This looks good, thanks!

PR #1020, opened 2017-10-15
Add SkeletonData cache support.

halx99:


badlogic:
Thanks for the PR! I'm in favor of not merging this though I'm afraid, as such a cache is application level logic and doesn't necessarily belong into the Spine runtime. I let @NathanSweet decide this though.

NathanSweet:
Agreed regarding application level logic. Thanks for sharing though!

PR #1021, opened 2017-10-17
Fixed: skinField property from SpineAttachment attribute

kleber-swf:
The `skinField` property from `SpineAttachment` attribute was being ignored because the `SpineAttachmentDrawer.PopulateMenu` was looking for the property name inside the `SerializableProperty` (which is a string).


pharan:
Thanks for reporting!
We were actually using a separate extension method with the signature `SerializedProperty.FindBaseOrSiblingProperty(string)` so it could handle both being an actual field of a MonoBehaviour or a part of a serialized class or struct field (It couldn't do that before and it was a really annoying limitation).

But overlooked applying it to skinField.

PR #1032, opened 2017-10-31
[Unity] Handle duplicate key

Abedron:
Error was after Apply in example project MixAndMatch

### DIFF
--- a/spine-unity/Assets/spine-unity/Modules/AttachmentTools/AttachmentTools.cs
+++ b/spine-unity/Assets/spine-unity/Modules/AttachmentTools/AttachmentTools.cs
@@ -585,11 +585,16 @@ namespace Spine.Unity.Modules.AttachmentTools {
 			if (output == null) {
 				Texture2D sourceTexture = ar.GetMainTexture();
 				Rect r = ar.GetUnityRect(sourceTexture.height);
-				int width = (int)r.width;
-				int height = (int)r.height;
-				output = new Texture2D(width, height, textureFormat, mipmaps);
+				int width = Mathf.Min(sourceTexture.width, (int)r.width);
+                int height = Mathf.Min(sourceTexture.height, (int)r.height);
+                output = new Texture2D(width, height, textureFormat, mipmaps);
 				output.name = ar.name;
-				Color[] pixelBuffer = sourceTexture.GetPixels((int)r.x, (int)r.y, width, height);
+                Color[] pixelBuffer = null;
+                int x = Mathf.Max(0, (int)r.x);
+                int y = Mathf.Max(0, (int)r.y);
+                
+                pixelBuffer = sourceTexture.GetPixels(x, y, width, height);
+
 				output.SetPixels(pixelBuffer);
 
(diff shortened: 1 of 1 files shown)


pharan:
What was the error after Apply in example? I'm not getting it.
Can you post the actual error?
This fix isn't quite right but I have some other suspicions of where it might actually be.
eg https://github.com/EsotericSoftware/spine-runtimes/commit/06a18da27be7a24f5583eb5c1c3ae16563200e5a

Abedron:
After run click to Apply
https://prnt.sc/h4diqp

Abedron:
I have next error with to the same method "Texture rectangle is out of bounds"

Fix for me:

```
Texture2D sourceTexture = ar.GetMainTexture();
Rect r = ar.GetUnityRect(sourceTexture.height);
int width = Mathf.Min(sourceTexture.width, (int)r.width);
int height = Mathf.Min(sourceTexture.height, (int)r.height);
output = new Texture2D(width, height, textureFormat, mipmaps);
output.name = ar.name;
Color[] pixelBuffer = null;
int x = Mathf.Max(0, (int)r.x);
int y = Mathf.Max(0, (int)r.y);
                
pixelBuffer = sourceTexture.GetPixels(x, y, width, height);

output.SetPixels(pixelBuffer);
```

pharan:
Does applying the change I linked fix things for you?
It **should** make it so it doesn't incorrectly enter into the `if (output == null)` block anymore.

The changes you made so far seem to address the symptom but not the cause. This is also pretty hot code (which is why the caching exists) so we hesitate to add unnecessary checks for obviously broken cases. This is not likely to be merged unless the conditions that cause it make sense.

Abedron:
I create reproducible project and send you. You may find a real problem.

Abedron:
New version spine-unity-3_6-2017-10-31 not have problem with apply, but still have problem with skin change.

Download and open `/Scenes/SpineProblem.unity`
https://drive.google.com/open?id=0B2dvOKWhu9WHYlRYZUtHZDBGcjA

After click "Hat" button get an error `Texture rectangle is out of bounds (0 + 801 > 497)`

pharan:
Thanks for sharing the project! I'll check it out.

pharan:
This seems to just be caused by your texture settings being incorrect.
You need to use Full Rect packing whenever you use Sprites to incorporate into Spine skeletons as RegionAttachments. Apart from not knowing what areas might be trimmed out, we also can't tell what other sprites are packed immediately outside.

See the recommended settings here: https://github.com/pharan/spine-unity-docs/raw/master/img/spine-runtimes-guide/spine-unity/mixandmatch-texture-inspector-options.png
It can also be found in the Mix-And-Match documentation. https://github.com/pharan/spine-unity-docs/blob/master/Mix-and-Match.md

We currently don't have the option for using the tight packing. We'll probably implement it soon. There have been recent major changes to Unity Sprite packing that prevented us from committing to the feature previously.

Abedron:
Thanx

PR #1034, opened 2017-11-03
fix warnings

minggo:


badlogic:
Thanks, what compiler and flags are you using to get the warnings?

minggo:
I think you can just use Xcode 9 to compile it. And i think it may exist on other Xcode versions. As you see, there are two types of warnings:

* declare shadow variables
* not explicit assign one type to another and may cause value lost, for example `unsigned long` -> int

minggo:
@badlogic any problem of this PR?

badlogic:
Sorry, no problem with the PR. I overlooked it :D

minggo:
Thanks

PR #1042, opened 2017-11-28
fix comment: CPP comment not allowed in C90

gutomaia:
Just a minor fix on a CPP comment, need to keep up with C90 standards!

```
/spine-runtimes/spine-c/spine-c/src/spine/AnimationState.c:183:1: error: C++ style comments are not allowed in ISO C90
 // These two functions are needed in the UE4 runtime, see #1037
 ^

```


badlogic:
Thanks!

badlogic:
Just wondering, what platform do you compile for via C90? Something embedded?

Issue #981, opened 2017-09-04
Investigate C++11 runtime


badlogic:
While our spine-c runtime is pretty good, it is still cumbersome to use from within C++. Ideally, we'd have a C++11 based runtime compatible with as many platforms as the C89 based runtime.

- [ ] Define platforms to support
- [ ] Check C++11 support of toolchains for each platform
- [ ] Come up with C++11 coding subset and standard that works on each platform
- [ ] Implement runtime is subset is "good enough" and actually a win over C89

ghost:
I'm currently integrating the spine-c runtime into my C++03 game framework now, which targets Windows 7+, Mac OS X 10.8+, iOS 8.0+, UWP 10, and Android 16+. I plan on leveraging rapidjson to handle file loading since it is already being used in my framework. I'm thinking I will be porting the whole c runtime into a C++ runtime for my use.

Just thought I would bring this up, though it sounds like you are wanting to target C++ 11 explicitly, I will be checking in on this issue periodically.

badlogic:
I'm not hellbound on needing C++11 features, C++03 may suffice. From what I read so far, C++03 may be the only thing that is also guaranteed to work on consoles. We don't have access to any console SDKs, so can't confirm this atm.

ghost:
Then in that case, I might have a PR for you pretty soon, since I plan on having my C++03 implementation completed here in a few days.

badlogic:
@sgowen that sounds great! To merge the PR and integrate your contribution into our repo, we need you to sign our [CLA](http://esotericsoftware.com/files/cla.txt) and have it sent to contact@esotericsoftware.com. Sorry for the red tape :/

ghost:
@badlogic No worries! We'll discuss once I get the PR ready.

badlogic:
Results of my investigation.

# Target platforms & Tool support
We aim at Windows, Linux, macOS, Android, iOS, WebAssembly as well as consoles. C++11 is supported on the first 5 platforms to the extend we'd need it for a spine-cpp platform. For consoles, the information is sparse, but from my research it appears that at least C++03 is supported.

# C++ Subset
I'm a strong believer in Branimir's [Orthodox C++](https://gist.github.com/bkaradzic/2e39896bc7d8c34e042b#file-orthodoxc-md). This will increase compatibility and guarantee that we don't fall into the "OMG, need to use all the shiny features" trap.

There are two areas that need some thought:

[...]

PR #1046, opened 2017-12-07
Spine-CPP Runtime

ghost:
Hey guys, sorry for the delay on this, I had intended to get this done much sooner, but it ended up being a lot of work. Anyways, here we are, let me know what you think!

A few notes:

1. I have the basic unit test running, but I didn’t get it configured the same way that the spine-c runtime is. Mine is literally just a simple test program; however, it does go through all the same steps as the C_InterfaceTestFixture and MemoryTestFixture classes, meaning that this C++ Runtime passes all the same tests as the C Runtime.

2. STL isn't used at all. Instead of std::vector or std::unordered_map, I wrote custom Vector and HashMap classes.

3. In all areas where new is used, it is only the placement new operator, not the typical default. This allows me to use the existing malloc/realloc/calloc/free overrides that spine-c was using. Also, I had implemented the Vector using C-style memory allocation, so it just sort of made sense to me to go this route. In my opinion, it really doesn't make a difference as long as you are careful to explicitly call the destructor, which I've mapped to a DESTROY macro.

4. The Json, SkeletonJson, and SkeletonBinary classes contain mostly borrowed code from their Spine-C Runtime counterparts.


ghost:
Adding in the MemoryTest now.

badlogic:
Hi! This is super cool! It will take some time for me to go through it and check it out. Please don't be mad if that happens after the new year! Really appreciate your effort, looking forward to exploring the code base.

ghost:
@badlogic No worries! Looking forward to it! In the meantime, have a Merry Christmas and a Happy New Year!

badlogic:
I took a quick looksy. I hope you won't be offended if we change the coding style to braces on same line :D. The RTTI implementation is neat! I saw a bit of caching going on here and there, need to check the memory management implications of that.

The bigger question for me is what we do with spine-c based runtimes. Should SFML/cocos2d-x/UE4 be updated to use spine-c? All of them are actually C++ based frameworks/engines, which require a minimum of C++11 at least in cocos2d-x and UE4 case. I could imagine we keep SFML spine-c based, and switch cocos2d-x/UE4 over to cpp. Or we keep two versions of each, though I'd find that 1) confusing and 2) a maintenance burden that's likely not worth it.

ghost:
Ahhh man, braces on the same line huh? Haha.

Yeah, literally all of the other runtimes are using braces on the same line, I suppose I just like having them on the next line so much that I didn't even think about it. I'll update it though.

Luckily, as far as the memory management is concerned, you'll be able to use the Kanji classes you are already familiar with to track all of the mallocs to get a feel for everything.

Hmmm, yeah, that's a good question. I wonder how much more portable the C Runtime is than this C++ Runtime, if at all? That would be interesting to know. Also, if there are any custom extensions of the C Runtime in use today that depend on its regular updates.

ghost:
Done.

Now I can't read the code... hahaha

I suppose a lot of my inspiration for braces on the next line comes from the libraries I use in my games: Box2D, glfw, libpng, and steamworks.

But then I also use rapidjson extensively, and it uses braces on the same line, so I guess it really just boils down to personal preference in the end.

badlogic:
Oh, wow, I didn't expect you to change the style! Thanks! Here's why we prefer braces on the same line: vertical screen space is limited. Braces on the same line eat up that space quickly, reducing the number of semantics bearing lines that can be viewed without scrolling.

ghost:
@badlogic Never thought about it that way, good point, makes sense! I imagine being able to view more without scrolling is even more so important on a codebase that gets lots of PRs.

mrpuzzler:
sgowen, great job! but why don't you use std::vector and other std:: parts? This is c++ and 2018 year! c++11 or c++14 already here!

jpoag:
My team uses an [AStyle ](http://astyle.sourceforge.net/astyle.html) configuration script to keep the source formatting for our projects the same.  Then I wrote a little script file that automatically formats all *.cpp/*.h files in a directory and then commits the changes to the repository.

Our teamcity server runs the script once a week, but I'll occasionally trigger the script manually.  If you work with badlogic, you should be able to get an official `config.astyle` configured.

For instance, the bracket style badlogic uses is called '[java style](http://astyle.sourceforge.net/astyle.html#_style=java)' in Astyle.

Example Config Script:
```
# style for all esoteric projects

--suffix=none # do not retain a backup of the original file

--style=java

--indent=spaces=4
--convert-tabs

--indent-switches
--pad-oper

--align-pointer=type
--align-reference=type

--max-instatement-indent=40

--keep-one-line-blocks
--keep-one-line-statements
```

Run:
`astyle --options=config.astyle -r ..\..\[source_dir]\*.cpp`

ghost:
@mrpuzzler here, let me link you to the original issue: https://github.com/EsotericSoftware/spine-runtimes/issues/981

@jpoag cool! I might have to give that a run on this PR. For my personal projects though, I'll probably keep using my own format (similar to what Box2D uses).

@badlogic Let me know how the PR is looking when you get a chance. I've been keeping it up-to-date with EsotericSoftware:3.6

mrpuzzler:
@sgowen, ok, I see why you don't want to use std (consoles). But you can make some typedefs inside extension.h ... so anyone can override std::string or std::vector with own implementations.

ghost:
@mrpuzzler Yeah, I totally could. I think at the time, when I was putting all of this together, it was just so much work, I was only focused on finishing up the port. I think once this PR gets merged in it would be cool to make spine-cpp more customizable.

badlogic:
Hey @sgowen, could you tell me up to which revision on which branch you've ported things? I need a baseline so I can port changes from the 3.7-beta branch to your runtime.

ghost:
@badlogic This PR is up-to-date with the latest spine-c on the 3.6 branch

spine-cpp/spine-cpp/include/spine/AtlasAttachmentLoader.h
```diff
@@ -0,0 +1,71 @@
...
+    
+    ///
+    /// An AttachmentLoader that configures attachments using texture regions from an Atlas.
+    /// See http://esotericsoftware.com/spine-loading-skeleton-data#JSON-and-binary-data about Loading Skeleton Data in the Spine Runtimes Guide.
+    ///
+    class AtlasAttachmentLoader : public AttachmentLoader {
+        RTTI_DECL;
+        
+    public:
+        AtlasAttachmentLoader(Vector<Atlas*>& inAtlasArray);
```
badlogic, line 50: Why does this take a `Vector<Atlas>` instead of just an `Atlas`?

ghost: This class is based on the c-sharp AtlasAttachmentLoader, which takes in an array of Atlas objects.

badlogic:
I've started investigating your PR. I've commented on one of the files here on GitHub. I assume you had a use case for this.

The bigger issue I have at the moment is that the placement new in combination with the custom memory allocation doesn't compile for me (Clang 9.0.0 on macOS X):

```
/Users/badlogic/workspaces/esotericsoftware/spine-runtimes/spine-cpp/spine-cpp-unit-tests/tests/C_InterfaceTestFixture.cpp:56:10: error: unknown type name 'skeletonJson'; did you mean 'SkeletonJson'?
    new (skeletonJson) SkeletonJson(atlasArray);
         ^~~~~~~~~~~~
         SkeletonJson
```

I've tried to rewire the allocation using the default C++ allocator, but that messes with the destructors and leaks like a sieve :)

`MemoryTestFixture.cpp` didn't compile at all, as it's using the spine-c API. Not a biggie, I'd rewire this myself.

The question is if we can modify the custom allocation scheme so it looks more like standard C++ allocation. I'm not a big fan of the use of macros in C++ for allocation. I'll see if I can come up with something.

Thanks so much for this PR! It's a great time safer for me to get the C++ runtime going! We've decided to move all engine runtimes using spine-c to spine-cpp except for SFML (which we keep as a reference implementation) and cocos2d-objc (which can't do spine-cpp). We'll include the SFML C++ runtime in the 3.7 release, with other runtimes added in the 3.8-beta. This will allow people to test the waters before jumping on the C++ train.

ghost:
@badlogic As mentioned in the PR description:

In all areas where new is used, it is only the placement new operator, not the typical default. This allows me to use the existing malloc/realloc/calloc/free overrides that spine-c was using. Also, I had implemented the Vector using C-style memory allocation, so it just sort of made sense to me to go this route. In my opinion, it really doesn't make a difference as long as you are careful to explicitly call the destructor, which I've mapped to a DESTROY macro.

As for the unit tests not compiling, it definitely works for me, also on macOS. Did you see the spine_unit_test.xcodeproj ?

But yes, I completely agree with you, I'm not a fan of using macros for memory allocation in C++ either, but I was't sure if people would be cool with losing the ability to use the Kanji utils for investing memory allocation, so I just kept things in-line with spine-c.

What we could do is just override the new and delete operators, but I've read that this introduces a whole other list of pitfalls. Of course... the simplest thing to do is just replace the NEW and DESTROY with the standard new and delete.

But what do you think?

ghost:
@badlogic Oh hey, I'm not using C_InterfaceTestFixture.cpp, maybe that's the reason. See main.cpp

badlogic:
@sgowen Hm, CPP_InterfaceTestFixture is empty when pulling the lastest from your branch. C_InterfaceTextFigure.cpp uses the C API, but the placement news don't compile. I'll poke at it and see if I can figure out an allocation scheme that is more C++-ish, while keeping memory leak detection in the unit tests.

badlogic:
@sgowen re:`Vector<AttachmentLoader>`, I see now that our C# runtime has this (@pharan manages that. In Kirk's voice: PHAAAARAAAAAAAAN). The reference implementation is the libGDX/Java one. I'll go through your port and will fix up any such discrepancies.

ghost:
screen shot 2018-01-29 at 11 08 18 am
@badlogic Yeah, I couldn't get the C_InterfaceTextFigure.cpp to work, so I just made 2 files: SimpleTest.h and MemoryTest.h, to do all of the same tests that spine-c was doing. I built/ran it just now to double-check and it worked fine.

ghost:
@badlogic re:Vector<AttachmentLoader> Ah! Whoops!

ghost:
@badlogic Sorry for the load of work I've just dumped on you >.< haha Let me know if there is any way I can help!

badlogic:
Hah, are you crazy? If anything, I'm the one who's sorry! This is a massive
amount of work. I'm happy I just have to tie up some lose ends. Will
comment on changes I make here.

ghost:
@badlogic Hahaha, and yes... a little... LOL

ghost:
@badlogic Oh did you see my comment above regarding the placement new operators? You need to be compiling main.cpp, SimpleTest.h, and MemoryTest.h, not the Test Fixture stuff.

badlogic:
@sgowen ah, I see it now. I'll have to move those tests over to the CMake system since we need to execute them on all supported platforms.

badlogic:
New memory management plan: each Spine class derrives from a base class called SpineObject (yeah...) which overrides new/delete. This new/delete will use the extension singleton for allocation. Internal code that allocates standard types, e.g. float arrays etc, will directly call into the singleton for memory. A bit more tedious on the internal side of things, but users can rely on the standard new/delete and are never exposed to macros.

I'm also ripping out the overly tedious unit test framework. With memory allocation tracking being easily implemented via the Spine extension, and cassert inclusion, it's easy to come up with a very simple test framework of our own without all the bagage.

badlogic:
I'm also going to replace the use of cassert inside the implementation with a method in the extensions.

ghost:
@badlogic I like the use of inheritance for achieving a better memory model! After all, that's what moving to C++ is all about right? 👍

ghost:
@badlogic So there shouldn't be any more need for malloc calls then right?

badlogic:
@sgowen re:malloc, only in some places where we instantiate PODs, e.g. float*, char*.

Vector does copy on insert and other operations. I'm not sure that's what we want. Spine's object graph is constructed in a way that an object's address indicates its identity. I might have to rewrite Vector and make it a bit less smart :)

badlogic:
I also just saw that a sine lookup table is used. We can't do that, as some calculations require full precision. On modern hardware, LUTs can also be slower than simply executing the corresponding CPU instruction.

badlogic:
Ok, the LUT is not your fault, I blame @pharan :D

ghost:
Whoops! So is the C# codebase out-of-date then, or does it just make more sense to use LUTs in that environment? I did think it was a little weird myself, but I didn't want to just assume that we didn't want them.

badlogic:
Not out of dead, it's just not the reference implementation. Once I'm done cleaning up I'll go through each LOC and compare to the reference implementation :)

badlogic:
I've opened an issue with the remaining tasks here https://github.com/EsotericSoftware/spine-runtimes/issues/1073

badlogic:
The use of references everywhere combined with assignment overloads etc. makes it really hard to make the code not leak. Did you port from spine-c or from spine-csharp? It looks like a port from spine-csharp. Some semantics have been lost in translation :/

I'm running into tons of memory issues which I just can't resolve. See 3.7-beta branch with the new simple test harness. I guess my C++ foo is just not good enough.

ghost:
I ported from both. And wow, there are memory leaks??? Makes me wonder if maybe we should just start this thing over using 3.7 beta as a base.

At this point I'm beginning to wonder if I've caused more harm than good!!!

badlogic:
Yeah, HashMap never let go of `Entry`, copy semantics in combination with references leads to double frees, etc. I've tried to clean up all of these, but it's quite tedious. I keep hacking at it.

badlogic:
Do you actually have this in production or in a prototype? I wonder if my memory allocation changes f'ed things up.

ghost:
@badlogic I threw it together into a quick prototype using C++/OpenGL, and it was working fine.

And then my artist and I ultimately decided we wanted to just use old-school sprites instead, and took Spine out of the equation, so I haven't seen spine-cpp in action since then.

I can totally re-take a look at this though, just let me know!

But I am hoping you can just fix the problems you are seeing by making changes to HashMap and Vector, since the actual file loading mechanics were all taken/adapted directly from spine-c.

Of course, I might be terribly oversimplifying things...

ghost:
@badlogic oh, one other thing! My prototype was using the C-style allocations/tracking. That stuff is pretty particular, since you have to call NEW followed by (placement new), and pair it up with a DESTROY.

Is it possible that through your refactoring of removing these C-Style allocations, that the callers of said allocations got broken?

Like, for example, if you changed the NEW to use C++ style allocation, but then left DESTROY calls in the code, that'd be bad

badlogic:
Nope, replacing the C-style allocation was actually OK. My issue is with code like this:

```
	{
		Vector<float> a;
		{
			Vector<float> b;
			b.reserve(10);
			b.setSize(10);
			b[0] = 123;
			b = a;
		}
		printf("%f\n", a[0]);
	}
```

I keep digging!

ghost:
@badlogic Setting a Vector equal to another should just call the = operator, and it should just work:

```C++
Vector& operator=(Vector& inVector) {
            if (this != &inVector) {
                clear();
                deallocate(_buffer);
                _buffer = NULL;
                
                _size = inVector._size;
                _capacity = inVector._capacity;
                
                if (_capacity > 0) {
                    _buffer = allocate(_capacity);
                    for (size_t i = 0; i < _size; ++i) {
                        construct(_buffer + i, inVector._buffer[i]);
                    }
                }
            }
            
            return *this;
        }
```

ghost:
@badlogic I guess what I don't understand is that spine-cpp passes all the same unit tests as spine-c, so how are we running into issues? Did you ever run the main.cpp / SimpleTest.h / MemoryTest.h ? These 3 files encompass all of the spine-c unit tests.

badlogic:
Ay, I changed that around to not do copy on assignement that way. I think I
dug my own grave :D Just needed confirmation from your end that you had it
all working.

ghost:
Oh I see. Do you just not like the operator= ? Or were you trying to make Vector more closely as just a glorified primitive array wrapper or something?

ghost:
Oh btw how different is 3.7-beta from 3.6 ?

badlogic:
Pimitive array wrapper. There was a lot of shuffling and copying data
around which we can't afford which can be replaced by address comparisons
instead. E.g. in SkeletonClipping teo float vectors would be compared by
element, which can add considerable overhead compared to a simple address
comparison.

I hope I didn't get the wrong impression across: I'm super happy you
contributed this! Just frustrated with my fuck ups.

ghost:
@badlogic no not at all! I feel bad that I did all this work (primarily for myself), and now I'm most likely not even going to be using it in the game I am working on, meaning my eyes aren't on it as much as they were before.

As far as all that overhead, you are totally right. What if we just added methods to Vector to just access the array directly?

I think the main issue with the data copying is directly a result of my influence by spine-csharp: for the data objects and just how all the data relates to each other in general

badlogic:
@sgowen Well, well, well, turns out I fluxored your beautiful Vector implementation to the point where i forgot to call a placement new and all shit broke lose. Fixed it up, but in the process destroyed the hashmap. I'll  have to replace the HashMap though, as it's default initializer would generate 64k Entry instances. Entry is 64 bytes on x86_64, so that's 4MB per HashMap :D

Now the next question is: shall we replace all uses of std::string with a custom string? The issue is that we can't fit std::string in our new allocation scheme (or the old one you setup), so a large amount of memory can't be managed by the user. Doesn't have to have all the fanciness of std::string I think. What do you say?

ghost:
Ouch, yeah 4MB hashmaps are no-bueno!

And hmm, sounds like we are talking about a const char* wrapper using our memory allocation hooks?

ghost:
And I'm super glad to hear you are progressing with all of this, makes me feel much better :)

badlogic:
Yeah, an immutable const char* wrapper would make the most sense imo. Or we simply use char*, though a wrapper would be nice for RAII.

ghost:
Yeah, that would definitely work.

Do you think there are there any other sneaky memory allocations happening that are dodging the Spine Memory Alloc macros?

badlogic:
Nope, everything else is covered. My problem is sneaky stuff like this https://github.com/EsotericSoftware/spine-runtimes/blob/3.7-beta/spine-cpp/spine-cpp/src/spine/SkeletonClipping.cpp#L283. In spine-c, that'd have been a pointer comparison, here it actually mows through both arrays, comparing every element. I'll have to check through each file and see what places like this need special treatment.

ghost:
haha, you could just delete the contents of the != operator, and replace it with a static_assert(false), that'd be a super quick way to know how many places we are doing Vector comparisons!

badlogic:
It's not just the comparisons though, the assignment operator is an even bigger problem. https://github.com/EsotericSoftware/spine-runtimes/blob/3.7-beta/spine-cpp/spine-cpp/src/spine/SkeletonClipping.cpp#L280 This makes an element wise copy every time (there are more places like this in the same method). The spine-c implementation of SkeletonClipping doesn't allocate or copy apart from potential resizing on elements inserts (and the strategy there is a classic "increase capacity by 2x"). Thankfully CLion can point out all the places where both operators are used, shouldn't be a big deal :)

ghost:
Ahhhh I see, well I hope this is all easy enough to undo. So basically, no copying at all unless absolutely necessary then right?

badlogic:
Pretty much, yes :) Shouldn't be to hard, might have to give up on the use of references here and there.

ghost:
I think I code way too defensively, and that's why we have all these references... >.<

badlogic:
Better defensive than dumb. Like me. Spending 2 days in the debugger and memchecker because I removed a placement new is not fun.

badlogic:
I'm  nearing completion of this. All the test now work as intended. There's just a few minor issues to fix up and the runtime is ready for primetime. Again, big thanks for all the work you've done!

ghost:
@badlogic This is exciting!!! Sweet!

PR #1053, opened 2017-12-22
old "fix warnings" exist some error, and this fix it

drelaptop:
`deform` was renamed to `deform2`, but some places where `deform` was quoted were not modified at the same time.

I think it's the reason you revert "fix warnings (#1034 )", now I revert you revert commit and fix this.


drelaptop:
This fix have been tested on win32, macOS && iOS by using cocos2d-x, all works well.

NathanSweet:
@badlogic What shall we do?

badlogic:
Obsolete because of C++ runtime.

Issue #1043, opened 2017-12-05
TwoColorPolygonBatch should implement Batch


badlogic:
See http://esotericsoftware.com/forum/Slot-getDarkColor-Tint-Black-9628

piotr-j:
Ive figured ill take a look at this, might need it in the future. Ive encountered one problem, Sprites. Since its kinda pointless without adding support for dark for other stuff, sprites verts no longer match. Also, another freaking batch...  any ideas for what, if anything, to do anout that?

badlogic:
Welp, no way around just manually transfering Sprite vertices to the batch
buffer, setting the dark color of each vertex to 0 :/

PR #1057, opened 2018-01-04
TwoColorPolygonBatch implements Batch interface, updated gdx to 1.9.8

piotr-j:
TwoColorPolygonBatch now implements Batch interface with support for dark tint for all draw types. 
Sprite vertices are transformed when submitted. 
Theres one issue, `SkeletonRenderer#draw(Batch)` and `SkeletonRenderer#draw(TwoColorPolygonBatch)` are now ambiguous. Im not certain whats the best way to solve that.

I had to update gdx for separate blend functions.

See #1043

### DIFF
--- a/spine-libgdx/spine-libgdx/pom.xml
+++ b/spine-libgdx/spine-libgdx/pom.xml
@@ -50,7 +50,7 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<gdx.version>1.9.6</gdx.version>		
+		<gdx.version>1.9.8</gdx.version>
 	</properties>
 
 	<build>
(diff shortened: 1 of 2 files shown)


badlogic:
Awesome, thanks! The ambigious draw methods aren't an issue really.

piotr-j:
Looks like i didnt quite get it right, missed one corner case. Drawing text. BitmapFontCache uses `TwoColorPolygonBatch#draw (Texture texture, float[] vertices, int offset, int count)` and if theres more then 1 character it breaks the batch. Current solution cant be easily extended. Perhaps a flag for automatic conversion enabled by default?

Something like this might do the trick https://gist.github.com/piotr-j/669d77fc9b385d12a9a40c7258bc8b58

badlogic:
Could you open a new issue?

PR #1089, opened 2018-03-20
Use UE4's memory management functions.

NuclearCookie:
This fixes unexpected behavior on certain platforms for which
UE4 specifically limits the maximum allocation size of malloc
in favor of using optimized allocation schemes.

### DIFF
--- a/spine-ue4/Plugins/SpinePlugin/Source/SpinePlugin/Private/SpinePlugin.cpp
+++ b/spine-ue4/Plugins/SpinePlugin/Source/SpinePlugin/Private/SpinePlugin.cpp
@@ -53,9 +53,11 @@ static void * SpineRealloc( void* ptr, size_t size ) {
 }
 
 void FSpinePlugin::StartupModule() {
+#if !UE_EDITOR
     _spSetMalloc( &SpineMalloc );
     _spSetRealloc( &SpineRealloc );
     _spSetFree( FMemory::Free );
+#endif
 }
 
 void FSpinePlugin::ShutdownModule() { }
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks for this. I tested it with our example scenes and it crashes when a `SkeletonRendererComponent` is destroyed, e.g. when a level is destroyed, or when the editor is closed.

Do you have  a repro case for me where the UE4 limit can be reproduced? I haven't found such a case yet.

NuclearCookie:
Hey Badlogic,
I experienced the same issues and marked the issue as [WIP] since. 
The issue happens because the the SpineEditorPlugin is instantiated before the SpinePlugin.
The EditorPlugin already constructs some spine objects ( using the standard malloc ) and we try to free it with our custom free. 

I tried to also use the `_spSetMalloc` etc from the SpineEditorPlugin, but I get unresolved external errors. Sadly, I don't fully grasp what is causing the link errors. My best guess is a missing dll export. I don't have enough experience ( and time ) to properly fix it. 

So for now, I don't use the custom memory functions if we are running editor mode and that fixes the issue.

A reproduction case will be quite difficult, as we experienced this issue on a closed source console. All sources are under NDA :(

On the platform we are using, system malloc only has 2Mb available. For the other allocations, a binned allocator is used. You could try to reproduce this by somehow limiting the amount of available space for malloc to 2Mb. No idea how to do that on a desktop platform!

badlogic:
Thanks for the great explanation! All I need to merge this :)

NuclearCookie:
If you know of a way to override the memory functions in the editor plugin, please do so! it will provide additional debug information like memory tracking tools.

badlogic:
I managed to fix up the linker errors in the editor plugin. The problem is that FMemory seems to use DLL specific allocators. Allocating something in DLL 1 and deallocating it in DLL2 is no bueno. I'll keep your last commit for the time being until I have some time to dive deeper into this.

NuclearCookie:
Great. Thanks for looking in to it!

PR #1096, opened 2018-04-10
Cocos2dx Skeleton culling

sp-miguel-ibero:
Spine animations can move the vertices of a skeletal mesh to arbitrary positions on the
screen. This causes a mismatch between the position of the cocos2dx node associated with a skeleton and the effective one on screen. Frustum culling requires the correct bounding shape, or at least a conservative one, of the skeleton in order to work correctly. Otherwise, it culls meshes that are actually visible on screen or, vice versa, keeps processing meshes entirely off screen, with the associated overhead.

The solution consists in computing the oriented bounding rectangle of all vertices of a skeleton each frame, and in using a new accurate 2D culling routine.


badlogic:
This looks great! I've only made 2 minor comments, nothing world chaning :)

Since this is a large change, could you please fill out, sign and send a photo the [CLA](http://esotericsoftware.com/files/cla.txt) to us? Cheers!

badlogic:
hey @sp-miguel-ibero, did you get a chance to check out my comments? I'm also afraid we can't merge this without the CLA.

sp-miguel-ibero:
Hi, sorry I didn't see the last comment. I sent the CLA by email the 12th of April and you answered that it was ok.

badlogic:
Hi! Thanks for the CLA. We have since rewritten the cocos2d-x runtime to use C++, see the 3.7-beta-cpp branch. Is there any chance you could rebase this onto that branch? Alternatively, could you maybe consider rebasing this to the 3.7-beta branch, where the cocos2d-x runtime is using spine-c?

sp-miguel-ibero:
Ok! We'll get to it. We'll first rebase to the 3.7-beta branch, and then take a look at porting the changes to C++ which may take some more time.

PR #1104, opened 2018-04-19
Fix code examples accessing internal fields

Xazen:
Use public properties instead of internal fields in code examples to allow spine to be placed in Plugins subfolder.


pharan:
You are correct. Thanks for pointing this out!

Issue #1117, opened 2018-05-24
[c] spAnimationState_apply optimization


ievgenmukhin:
https://github.com/EsotericSoftware/spine-runtimes/blob/d11ebfededd3ae285227f6cd1f93d3370dc80ff7/spine-c/spine-c/src/spine/Skin.c#L74

spAttachment* spSkin_getAttachment (const spSkin* self, int slotIndex, const char* name) which is called for each timeline in spAnimationState_apply consumes most of CPU time during animation state applying, especially for spine files with large amount of images. We have spine file with ~1000 images for pre-rendered animations, some of them used in more than one bone, resulting in ~8K entries in _spSkin. With such file we have very low framerate on older hardware.

Adding hashtable to _spSkin for fast Entry lookup gives us ~40x performance gain on such large spine animations.

badlogic:
Ay, `spSkin_getAttachment` performs a linear search through the slot/name entries. Replacing that with a map definitely gives a good speed up. If you have made this change already, we are happy to merge a pull request. Otherwise I'll look into fixing it up asap.

ievgenmukhin:
Hi, I have implemented this as temporary fix using 3rd party code from public domain for string hashing (murmur hash), so I'm not sure if it can be merged to your code base. I can update my fix with string hashing which is good for you, and create pull request if you want.

badlogic:
Murmur should be fine license wise. Happy to merge your PR!

ievgenmukhin:
Did you have a chance to check my PR? https://github.com/EsotericSoftware/spine-runtimes/pull/1118

badlogic:
I've commented on your PR! Sorry for the delay. Let's continue tracking this in the PR, closing this out.

PR #1118, opened 2018-05-28
[c] Optimized attachment lookup in skin. See #1117.

ievgenmukhin:


badlogic:
Sorry I haven't had time to merge this yet. Could I bother you to send us a scan or photo of the [CLA](http://esotericsoftware.com/files/cla.txt)? Otherwise we can't merge your PR. Sorry for the inconvenience.

While the solution is already quite simple (great job!), I wonder if we can make it even simpler. What if we bucket by slot index instead of calculating a murmur hash? That way, we could get rid of the hashing entirely, and just have an array of `_SkinHashTableEntry` which is indexed by the slot index instead of the murmur hash. My guess is that locality would actually be better.

If you want to make and test this change for your use case that'd be great, as we can directly compare performance to your murmur hash solution. Otherwise I'll try to get to this sometime this week.

ievgenmukhin:
I've update PR with your suggestions. Actually hash table distribution is better with murmur hash, but search works much slower with it (up to 2x on small spine files), probably because of time required for hash calculation. I got same search speed on large file with 8000 skin entries when hashtable size was increased to 1000 elements. I assume that with even more complex spine files and larger hashtable sizes murmur hash will start to work faster than your approach, but for most cases slot index is better. 
I'll send you CLA via email.

badlogic:
Thanks a lot, great work!

Issue #1119, opened 2018-05-29
[ts] Events keep firing during mix-out


badlogic:
See http://esotericsoftware.com/forum/TS-Multiple-events-fired-10395

NathanSweet:
Fix ready for porting: 2a7fd54224e53d73ae303bed480bdac10bacd838

badlogic:
Ported to all runtimes.

smaren:
@badlogic I see you have closed this issue, it has however not been fixed for the 3.6 branch (at least for the TS version)

badlogic:
I backported the changes to all 3.6 runtimes.

PR #1120, opened 2018-05-30
Fixes: Events keep firing during mix-out #1119

smaren:
Fixes issue #1119 @badlogic


NathanSweet:
Great, thanks for the fix! AnimationState is a beast. Let's use #1119 to track porting this to all the runtimes.

PR #1128, opened 2018-06-22
Fixed illogical comparison warning

aram-ahak:
Please, fix this (Illogical comparison with undefined.  Only untyped variables (or variables of type *) can be undefined) warning.


badlogic:
Cheers!

PR #1137, opened 2018-07-05
Add animation speed option

kit494way:


pharan:
The AnimationState class has a timeScale variable. You can probably use that.

https://github.com/EsotericSoftware/spine-runtimes/blob/71c0887dab46dabdcb5bb2232407862f9863ee27/spine-ts/core/src/AnimationState.ts#L47

kit494way:
Thank you for your information. It is very helpful for me.

PR #1165, opened 2018-08-28
Expose USpineAtlasAsset and USpineSkeletonDataAsset types for Unreal 4 blueprints.

npruehs:
Currently, it is not possible to set the current atlas and skeleton data of a USpineSkeletonAnimationComponent through blueprints.

Adding the BlueprintType property to the UCLASS definition resolves this issue.


npruehs:
Signed CLA follows!

badlogic:
Thank you!

PR #1167, opened 2018-08-29
Expose collision generation for skeleons for Unreal 4 blueprints.

npruehs:
Following up on the following discussions, automatic generation for collision data for skeletons in Unreal 4 has been disabled by default:

http://ar.esotericsoftware.com/forum/UE4-Spine-Runtime-Mesh-Component-performance-9873
http://esotericsoftware.com/forum/UE4-Performance-problem-example-project-current-branch-9482

Exposing the flag to blueprints allows us to decide for every skeleton whether to generate this data, or not.


npruehs:
Signed CLA follows!

badlogic:
This is a sensible middle ground. Thanks for the contribution!

PR #1170, opened 2018-09-05
Fix null refrence when accessing AnimationState of SkeletonGraphics i…

fshahinfar1:
…n build version. this issue is caused because of SkeletonGraphics not being initialized.


pharan:
Hello!

If you're instantiating SkeletonGraphic components or need to otherwise access them during the Awake phase, I'd recommend calling `skeletonGraphic.Initialize(false)` to ensure the skeletonGraphic is initialized before doing anything with it.
Otherwise, Unity documentation recommends [not doing any cross-component communication beyond just getting references to the components themselves in Awake, and just do those in Start](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Awake.html).

We're not changing the behavior of the 3.6 runtimes at this point.
Thanks for the PR! We'll consider what changes will be done in this context for 3.7

PR #1175, opened 2018-09-15
[UE4] UBoneDriverComponent and UBoneFollowerComponent as USceneComponent

przemyslawlis:
I would like to propose a change for both `UBoneDriverComponent` and `UBoneFollowerComponent `to be `USceneComponents `instead `UActorComponents`. We could then use component transform instead of owning actor transform which is more flexible and opens new options.

This behavior is optional, and can be switched on by setting `UseComponentTransform `property to true. Default value is set to false, which means that by default the behavior is unchanged from what it is now.


badlogic:
Thanks! Sorry for the delay.

PR #1196, opened 2018-10-15
Feature/new skeleton culling

sp-stefano-lanza:
I've ported https://github.com/EsotericSoftware/spine-runtimes/pull/1096 to the 3.7-beta branch. The only noteworthy change is the addition of early outs in loops when the alpha channel of slots, skeletons or attachments is zero, as they'd be invisible.


badlogic:
Thanks, looks and works great! I'm closing out the old PR as well.

PR #1223, opened 2018-12-07
Fix for CS1032 in 2018.3 Unity

23doors:
Getting CS1032 error under 2018.3. Apparently define needs to be done before any `using` declarations. https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1032


23doors:
I see @pharan fixed in on his own in https://github.com/EsotericSoftware/spine-runtimes/commit/044f0a764fb68590d6f0e35158acb4aa1c8f2d3a . Fair enough although could have just check PRs every once in a while...

PR #1236, opened 2018-12-28
[unity] Support Prefab Mode of Unity 2018.3 or newer

OpenLevel:
[ExecuteInEditMode] to [ExecuteAlways]
As a safety precaution, scripts with the [ExecuteInEditMode] attribute cause the Editor to exit Prefab Mode when going into Play Mode, if any script on the open Prefab has that attribute. For details on how to make such scripts compatible with Prefab Mode, see the Scripting Reference pages for ExecuteInEditMode and the new ExecuteAlways attribute.

in Editor, new GameObject to ObjectFactory.CreateGameObject
The Editor tooling that creates GameObjects should use ObjectFactory.CreateGameObject to make sure that the GameObject ends up in the Prefab Scene if a Prefab Asset is currently being edited in Prefab Mode.


OpenLevel:
Before the merge,
some '.vs' folder binary files should be deleted.
spaces should be changed into tabs.
Sorry.

pharan:
Interesting.
Thank you for this contribution.

Apart from those other extraneous changes you mentioned, we are also currently preferring the explicit "NEW_PREFAB_SYSTEM" define to make it clear what system the enclosed code is handling. We probably won't merge those changes as is.

pharan:
Instead of having the define per file, we're handling the new ObjectFactory API by centralizing editor GameObject instantiation in EditorInstantiation class.
https://github.com/EsotericSoftware/spine-runtimes/commit/40f5755fe0ca849431cc1cc46af0b005c97ca2a3

OpenLevel:
I completely agree with you! I agonized about the define. Your way is better. Keep improve your product! It is very helpful.

OpenLevel:
Also it slightly defferent tho, I think you can intergrate ExecuteInEditMode and ExecuteAlways as the same way. Thanks.

pharan:
`[ExecuteAlways]` was added: https://github.com/EsotericSoftware/spine-runtimes/commit/66689b3800bd19b72a48c10ead278434b11ed97f

I would squash and merge this to keep it on record but I don't know how to discard those extra files in the process without access to your branch. I'll keep the PR open a little longer in case the others know.

@NathanSweet also said you should sign this in that event: http://esotericsoftware.com/files/cla.txt
Though personally, this PR doesn't involve actual logic, just API changes, so I don't know.

Again, thank you very much for your contribution!

PR #1251, opened 2019-01-14
- (UE4) Changes meant to support PaperZD or any other external plugin…

heavybullets:
- (UE4) Changes meant to support PaperZD or any other external plugins which
    need to interface directly with the renderer.
- (UE4) Added skeleton animation caching on SkeletonRenderer to avoid
per tick searches which are costly. The skeleton animation component can
be setup externally.
- (UE4) Added bAutoPlays members to AnimationComponent, so the skeleton
doesn't update automatically if its needed.
- (UE4) Added SetPlaybackTime method to AnimationComponent.


badlogic:
Merged manually. Note that the auto play flag is set to `true` by default as to not break existing projects.

PR #1293, opened 2019-03-06
[cpp] Fixes #1286 TrackEntry now resets renderer object

### LINKED ISSUE #1286: [cocos2dx] Crash when setting a track complete listener
See http://esotericsoftware.com/forum/cocos2dx-Crash-when-setting-a-track-complete-listener-11570

SometimesMage:
Found that the TrackEntry properly deletes the RendererObject (at least in the cocos2d-x code) but never sets that pointer back to NULL which leads to problems later since track entries are pooled (again, at least in cocos2d-x). This patch will set the RendererObject back NULL when a TrackEntry is reset.


badlogic:
Cheers! And thanks for also proactively sending the CLA.

PR #1298, opened 2019-03-08
[ts] Make let not var

ivanpopelyshev:
subj. 

There was only one instance of `var` inside if, and I put two `let`s there, in `Animation.ts`


badlogic:
Cheers!

PR #1299, opened 2019-03-08
[ts] I gift you the second equals sign

ivanpopelyshev:
Need to check the other runtimes too.

UPD. OK, only [ts] has it.


badlogic:
Oh no... Cheers!

PR #1307, opened 2019-03-13
[cpp] Making the callbacks also object oriented

AlvaroEstradaDev:
The current callback system only allows for standalone functions to be passed, so you need to use global variables in order to make the callback interactable with the rest of the code. I added an abstract listener class in order to make the code more flexible when making callbacks. It's retro-compatible, so the legacy code wouldn't be compromised.

### DIFF
--- a/spine-cpp/spine-cpp/include/spine/AnimationState.h
+++ b/spine-cpp/spine-cpp/include/spine/AnimationState.h
@@ -58,6 +58,17 @@ namespace spine {
     class RotateTimeline;
     
     typedef void (*AnimationStateListener) (AnimationState* state, EventType type, TrackEntry* entry, Event* event);
+
+	/// Abstract class to inherit from to create a callback object
+	class SP_API AnimationStateListenerClass
+	{
+	public:
+		AnimationStateListenerClass() = default;
+		virtual ~AnimationStateListenerClass() = default;
+	public:
+		/// The callback function to be called
+		virtual void callback(AnimationState* state, EventType type, TrackEntry* entry, Event* event) = 0;
+	};
     
     /// State for the playback of an animation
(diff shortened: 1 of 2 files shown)


badlogic:
Couldn't you achieve the same by using a capturing lambda? That would not require breaking the API.

AlvaroEstradaDev:
My main issue came from trying to access private methods of multiple classes, while I could access the private methods of the class that created the lambda, it was a different issue accessing others. The functor class could be befriended by multiple classes, allowing the functionality I'm looking for.

AlvaroEstradaDev:
Writing that last comment made me realize that I could just pass my functor object to the lambda and avoid this whole issue.

AlvaroEstradaDev:
I did a test and found out that a capturing lambda cannot be converted to a conventional pointer function. Quoting the[ c++11 draft](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf): "The closure type for a lambda-expression with no lambda-capture has a public non-virtual non-explicit const conversion function to pointer to function having the same parameter and return types as the closure type’s function call operator. The value returned by this conversion function shall be the address of a function that, when invoked, has the same effect as invoking the closure type’s function call operator."
A functor abstract class is still the best option for flexibility.

badlogic:
Ah, right. I guess we can expand the API like you proposed in your PR.

However, when a function callback is set, the listener object should be cleared, and vice versa. At the moment, the object gets priority over a function callback, even if both are set. We only ever want one callback (function or object) to be set on the `AnimationState` or `TrackEntry`.

Could you amend your PR accordingly?

AlvaroEstradaDev:
Just made the update to only consider one listener at a time

badlogic:
Cheers! I've manually merged this PR with some clean-up modifications.

Issue #1047, opened 2017-12-11
[ue4] Add UMG support


badlogic:
Once this is implemented, use it in the assets editor, see https://github.com/EsotericSoftware/spine-runtimes/issues/1291

sumday:
How about this issue now?

Rainfarer:
Are there any news regarding this one?

motifgames:
We are waiting this feature 1 year. I want to know release schedule or roadmap. It is VERY IMPORTANT about Our Team and Project.

GitHub-NetFly:
There is no progress yet?

GitHub-NetFly:
Is this problem reopened? I'm still working on the problem that the Basic renderer can't render mesh..Once there is a result... I will continue to discuss it with you here.

GitHub-NetFly:
In communication with Epic engineers, I have made some progress on this matter, the effect is as follows:
![1](https://user-images.githubusercontent.com/6991952/54423124-e2384000-474a-11e9-86df-fb9558c70a52.png)

![2](https://user-images.githubusercontent.com/6991952/54423133-e6fcf400-474a-11e9-9e06-bc3e6c0aef79.png)


There are probably two problems. 1. The MaterialDomain used on the UI should use UserInterface, which may result in a UI version. All Blend types need to recreate a UI material version.
![334575-materialdomain](https://user-images.githubusercontent.com/6991952/54325163-c09c6300-463b-11e9-99ae-72541d94570e.png)

2. The coordinates of the vertices Y and Z need to be interchanged. In addition, the approach I take here is that Y takes a negative value.
![334576-vertadd](https://user-images.githubusercontent.com/6991952/54325195-e6296c80-463b-11e9-836a-6777faba875e.png)

Then re-map the vertices to the corresponding position according to the size to be drawn.
![334577-remapvert](https://user-images.githubusercontent.com/6991952/54325214-03f6d180-463c-11e9-9b71-0233d201fc00.png)

Related code：
[code](https://drive.google.com/open?id=1yFlW-E5njpMU5Afq-Yw3ReWK_j2EyUnH)

The code is for functional reference only and is not recommended for formal production use. It is not optimized and organized.

[...]

PR #1311, opened 2019-03-21
[ue4] Fix Widget rendering error

GitHub-NetFly:
I have made some progress on this matter.
For details, please refer to:
> https://github.com/EsotericSoftware/spine-runtimes/issues/1047

The code is for functional reference only and is not recommended for formal production use. It is not optimized and organized.

The principle of this rendering is to perform 2D skeletal animation in Spine's library, then get the position of the new vertices on each Slot, and then use these vertices to generate ProceduralMesh for rendering, one Slot corresponds to a Section. For Slate, it is to follow the MeshWidget approach, update the Mesh vertex data, and then draw the faces of these vertices in the OnPaint process..


GitHub-NetFly:
In addition, I found that strange effects appear at runtime, the effect is as follows:


Solution: Modify the texture properties


Or check the never stream

badlogic:
I'm working on this at the moment, but I have a problem I do not know how to resolve. When editing a widget with the Spine widget component, the Spine skeleton flickers like crazy. I tried to debug this, but i'm at a loss at why the flicker occurs. Any ideas?

badlogic:
I have now fixed up our existing widget code with some inspiration from your pull request in the 3.8-beta branch. I've also added methods for animating the widget. The widget `Tick` method needs to be called as part of a blueprint by the user of the widget.

badlogic:
@GitHub-NetFly any idea why the tick and construct events wouldn't trigger when inspecting a widget?

GitHub-NetFly:
> @GitHub-NetFly any idea why the tick and construct events wouldn't trigger when inspecting a widget?

Sorry, my English ability is limited..

Tick and construct should be fine.

Now outside, without looking at the code, if I remember correctly, the animation should be a transform of the frame of the slot passed in the plugin, and then according to the newly generated vertex position and uv information of the slot, the uv should be unchanged. Then, draw again with the same material.

Because the transform of each slot is not the same, so I remember it is correct, although it is a drawcall, the vertex position on the slate is to be regenerated, I felt that I can optimize, but did not continue to do it.

At present, tick should be once every tick, the material is unchanged, uv is unchanged, but the actual vertex position changes, the vertex position should be based on the tick time, the specific position of the slot, generate four new 2d positions, if you want to optimize, Should be the 2d transform of all slots into a uniformbuffer, and then according to the index of the slot to determine how to change the vertex position, I remember I said that we have problems when generating vertices, the coordinates of several axes interchanged, it is a Where the frame data is generated, theoretically every vertex of each frame is updated, and the slot will give new data every frame.. So I think the problem should be here?

Problem code:
QQ截图20190330140213

GitHub-NetFly:
> I'm working on this at the moment, but I have a problem I do not know how to resolve. When editing a widget with the Spine widget component, the Spine skeleton flickers like crazy. I tried to debug this, but i'm at a loss at why the flicker occurs. Any ideas?

I found that there is no problem in the case of DebugGame, I guess it is caused by the internal implementation of TArray.

badlogic:
All the flickering and related rendering issues have been resolved. The only open issue is that the tick events don't trigger, which I'm trying to resolve today. The `Test` content folder contains a UMG widget that I use for testing (on the 3.8-beta branch!). If you find time to take a look, let me know if you have an idea why tick doesn't trigger.

GitHub-NetFly:
> All the flickering and related rendering issues have been resolved. The only open issue is that the tick events don't trigger, which I'm trying to resolve today. The `Test` content folder contains a UMG widget that I use for testing (on the 3.8-beta branch!). If you find time to take a look, let me know if you have an idea why tick doesn't trigger.

I found out that DeltaTime is 0 through Debug. After I call it in BP, the animation can be played normally. But he is random. In some cases, he will render failed. I suspect there is a relationship with the vertex data.The code in the animation section should also have problems. When Tick is activated, the engine will crash.
QQ图片20190401175906

QQ图片20190402111005

It works fine in DebugGameEditor mode

QQ截图20190402110448
After packaging, the game will also crash..

Another problem is that in the editor mode, the Spine image will fail to render... it can be rendered normally under the engine window. I suspect this is a Layer problem, or it is handled separately under the design view.

GitHub-NetFly:
> All the flickering and related rendering issues have been resolved. The only open issue is that the tick events don't trigger, which I'm trying to resolve today. The `Test` content folder contains a UMG widget that I use for testing (on the 3.8-beta branch!). If you find time to take a look, let me know if you have an idea why tick doesn't trigger.

At present, it seems that Tick is triggered normally, but the value of DeltaTime in BP is always 0, so it feels that it is not triggered. After the Tick is triggered normally, the engine will crash, and packaging into exe will also crash.. But I am in DebugGameEditor mode. It will run normally.. I guess there should be a problem with the code in the animation part. The vertex processing of the rendering part should also be problematic.. Because I modified the rendering code before, I reversed the axis of Y and Z. .Y took a negative value.. I haven't had too much time to go deeper to see the code. After the work is over, I will continue to follow up on this question:

badlogic:
@GitHub-NetFly So, I managed to fix all this up, but I can not reproduce the crash of rendering failures you reported. Could you please send me a step by step instruction how I can reproduce the crash and rendering errors you see?

There is one issue left to resolve: the skeleton will always try to fill the widget size, which looks funny. Maybe you are referring to that?

GitHub-NetFly:
> @GitHub-NetFly So, I managed to fix all this up, but I can not reproduce the crash of rendering failures you reported. Could you please send me a step by step instruction how I can reproduce the crash and rendering errors you see?
> 
> There is one issue left to resolve: the skeleton will always try to fill the widget size, which looks funny. Maybe you are referring to that?

The problem of the crash, I explained in detail in the forum, can refer to here:
[unreal widget animation](http://zh.esotericsoftware.com/forum/UE4-Unreal-Widget-Animation-11660)

Widget Size problem, I suspect that there is a connection with the rendering failure, I have no clear way to trigger it.

badlogic:
Cheers, i was able to reproduce the mystery crashes. They were related to multi-cast delegates as well as a bad cast. The crash should be fixed now.

I can also reliably reproduce the issue of the meshes randomly disappearing. Not sure what the solution to that is yet. I'll figure it out.

Thanks for all your help with this! I'm closing the PR now. Let's continue the discussion in the issue #1047

PR #1342, opened 2019-04-27
casting exception when use custom atlas asset

lolisia:
casting exception when use custom atlas asset


HaraldCsaszar:
Thanks very much for the bug report and bugfix. It is now integrated in 3.7 and 3.8-beta branches and will be included in the next unitypackages.

PR #1360, opened 2019-05-22
expandToIndex should not subtract its current length

smaren:
If the length is first expanded to 2 and then to 5, the result end result is a list with length 3. That is not working as intended


smaren:
In fact, this seems to be the way it works in libgdx also, so it should probably be fixed in other places also

badlogic:
Cheers, I've fixed this locally. I also went through the other runtimes. I checked through all the other runtimes, and their way of handling the tracks array expansion is correct. Thanks for reporting!

PR #1374, opened 2019-06-05
[unity] Fix ps4 shader compile error

chaixl:
Use SV_Target to fix shader compile error on PS4 platform.


HaraldCsaszar:
Thanks for the issue report and fix!
All fragment shader outputs on all Spine shaders are now corrected, both on 3.7 and 3.8-beta branches.

PR #1395, opened 2019-06-24
[unity] Check more accurately whether a json is for Spine or not

marcusalanen:
Do not assume a json file with contents like '{ "skeleton": ...} '
means that the file is for Spine. Instead try a little harder and also
seek out the spine key under skeleton.

This came up in our game development. (Some enemies are skeletons :)


HaraldCsaszar:
This has been resolved in another way in the meantime. Thanks for reporting!

Issue #1397, opened 2019-06-29
[unity] Support for moving/renaming sprite texture files (destructive reimport)


leonidumanskiy:
If sprite texture file is renamed or moved to another directory, next re-import will destroy asset material and leave atlas asset in the broken state, even if nothing changed (i.e. project is being imported first time).

leonidumanskiy:
Side notes:

This breaks different workflows. 
For example, one would want to move textures to use Unity built-in asset bundle variants:

_character_skeleton.assetbundle_:
```
/character/character_Atlas.asset
/character/character_Material.mat
/character/...
```

_character_textures.assetbundle.HD_:
```
/character_textures/hd/character.png
```

_character_textures.assetbundle.LD_:
```
/character_textures/ld/character.png

```
_character_textures.assetbundle.SD_:
```
/character_textures/sd/character.png
```

HaraldCsaszar:
Thanks for reporting and sorry for the long delay! This has finally been fixed.

PR #1398, opened 2019-06-29
[unity] Fixed issue #1397 when renaming or moving sprite texture would result in next re-import destroying material and breaking atlas asset

leonidumanskiy:
Fixed issue #1397 when renaming or moving sprite texture would result in next re-import destroying material and breaking atlas asset.

AssetUtility will now attempt to keep the original texture of the material, if texture is not found during re-import but material is present.


HaraldCsaszar:
Thanks for reporting and sorry for the long delay! This has finally been fixed.

PR #1402, opened 2019-07-05
better for lua binding

codetypess:


badlogic:
I'm afraid we have to support C++03 and can't use a c++11 stdlib API.

PR #1404, opened 2019-07-10
define a macro to use std function

codetypess:
recently, i design a new lua binding for cocos2dx, i found that, std function is easy to generate lua bindings


badlogic:
Good idea! Happy to merge this. Could you retarget it at the 3.8-beta branch? We are about to release 3.8 in a week.

badlogic:
If added this manually to the 3.8-beta branch. Thanks for the suggestion!

PR #1408, opened 2019-07-12
missed implement for setSoftness and getSoftness

codetypess:


badlogic:
Thanks, I've resolved this manually.

PR #1419, opened 2019-07-25
INITIAL_WORLD_VERTICES_LENGTH can now be predefined by projects depending their needs 

benjaprieto:
Small modification, so INITIAL_WORLD_VERTICES_LENGHT can be predefined in projects with other value if needed.

Changed INITIAL_WORLD_VERTICES_LENGHT into SPINE_INITIAL_WORLD_VERTICES so it hints that is SPINE library related


badlogic:
Could you give me a use case for this? The memory block is shared across renderers and unlikely to be a memory hog unless your skeleton attachments have a massive vertex count.

badlogic:
Will re-open if you can give us a good idea why this is beneficial.

PR #1427, opened 2019-08-05
FNA Runtime

Predator75:
FNA is a reimplementation of the Microsoft XNA Game Studio 4.0 Refresh libraries.
FNA is primarily developed by video game porter Ethan Lee, who has shipped more than three dozen ports of XNA games using the exact branch that you see on GitHub today.
it can run on win,mac,linux and same console, it supports x86/x64
https://fna-xna.github.io/

FNA is fully supported by Steam.

I converted the spine xna runtimes to make them work with fna, it is very easy.
post on forum: http://it.esotericsoftware.com/forum/FNA-runtime-12301

I attach an archive with fna runtime
and I link a fully working test project: https://www.eprime.it/download/spine_test_con_sprite.zip (sorry we work in vbnet ;) )


badlogic:
Thanks, I had a look through the files. Could you tell me why you duplicated the files from `spine-xna` such as `MeshBatcher.cs`? FNA is a drop-in replacement for XNA, correct? As such, it shouldn't be necessary do make any modifications to these files.

Predator75:
Hi badlogic,
I have duplicated the files because it is the first time that I use a pull to github, and I was wrong to upload.
Exactly, there is no need to change anything, just to reference the fna library instead of xna, and remove the pipecontent that is not is no longer used.
All my projects with Spine I use exclusively in FNA runtime.
The goal is to preserve the XNA game code, with an incredible focus on accuracy. The changes that fna code need is zero or very very minimal.
FNA supports Windows, macOS, and GNU/Linux with a single assembly file, and has additional support for iOS, tvOS, Xbox One, and Nintendo Switch.

badlogic:
Ahh, all that makes sense. I think it makes the most sense to link to this issue here on our runtimes page so people know that FNA is supported as well. I'll do that in a bit! Thanks for your contribution!

PR #1429, opened 2019-08-05
INITIAL_WORLD_VERTICES_LENGTH can now be predefined by projects depending their needs

benjaprieto:
Changed INITIAL_WORLD_VERTICES_LENGTH to SPINE_INITIAL_WORLD_VERTICE_LENGTH, and the way it is declared, to be pre-declared with a different value if need by the project

### DIFF
--- a/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
+++ b/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
@@ -35,7 +35,10 @@
 #include <spine/AttachmentVertices.h>
 #include <algorithm>
 
-#define INITIAL_WORLD_VERTICES_LENGTH 1000
+#ifndef SPINE_INITIAL_WORLD_VERTICES_LENGTH
+#define SPINE_INITIAL_WORLD_VERTICES_LENGTH 1000
+#endif
+
 // Used for transforming attachments for bounding boxes & debug rendering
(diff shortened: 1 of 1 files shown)


badlogic:
As I said on your other PR, what is the purpose of this? What problem are you trying to solve?

benjaprieto:
> As I said on your other PR, what is the purpose of this? What problem are you trying to solve?

Even after the number was increased, this has not been enough for our project, which will crash if we do not increase it to a certain number

The proposed solution allow to give us the liberty to override the value, without affecting the values of the original project

Which other PR? I thought I had made the PR before, but was not able to find it, so I created this again (thought I might have got confused about creating the previous one)

badlogic:
The code should never crash. The memory block will be re-allocated if it doesn't provide enough space for the vertices of an attachment. This function is called in appropriate places to ensure that:

https://github.com/EsotericSoftware/spine-runtimes/blob/1d1d6e01d6a0a35371dc5fd6130ee14a9c7ab507/spine-cocos2dx/src/spine/SkeletonRenderer.cpp#L46

Which is why I'm confused why this PR is needed. Could you send me a reproduction sample that crashes reliably? I'd rather fix the underlying issue than work around a bug like this.

Could you open an issue with the repro case? I'll try to fix it asap then. Closing this out in the meantime.

PR #1431, opened 2019-08-06
Support for Gradle building, installing, and jitpack

Tom-Ski:
Allows for jitpack and local Gradle building which is handy as maven is not being deployed. No additional work required.

Minor changes:
gitIgnore update for IDEA and Gradle imports.


badlogic:
Btw, what do you mean with "Maven is not being deployed"? We release snapshots of the beta branches and occasionally release release versions.

Tom-Ski:
When I latest checked the latest release was January to central. And regrading snapshots I think they may have been down for a few months, April seems to be the latest there on the 3.8 builds. 
 But as there is no history I can't confirm this :(

https://mvnrepository.com/artifact/com.esotericsoftware.spine/spine-libgdx

Handy to have this anyway, we've used some internal patches to the runtimes whilst waiting for spine to update (to new LibGDX version for example), and jitpack makes this trivial to swap between the two.

badlogic:
Well, I blame you that we now also have Gradle in the repo. Thanks, I hate
it :D

Tom-Ski:
I noticed you already had npm, so it couldnt be any worse

PR #1436, opened 2019-08-08
Support for 3.6 and 3.7 format load in C runtime

vegorovinno:
This patch allows 3.8 C runtime to load files exported from 3.6.* and 3.7.* editor.

Additionally, it prevents the C runtime to load older and future file versions (original C runtime doesn't check the version and crashes on mismatch)

### DIFF
--- a/spine-c/spine-c/src/spine/SkeletonBinary.c
+++ b/spine-c/spine-c/src/spine/SkeletonBinary.c
@@ -34,6 +34,9 @@
 #include <spine/Animation.h>
 #include <spine/Array.h>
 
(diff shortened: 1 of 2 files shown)


badlogic:
I'm sorry, we will not merge this PR. Support of different runtime versions in the same binary is not possible. The behaviour of the implementation changes between versions. Even if you can load files of older Spine Editor versions, they will not behave as they did with the Spine Runtimes version matching the editor.

vegorovinno:
You should still add a version check to avoid hard crashes.

PR #1437, opened 2019-08-08
[c] Only load files that are created by matching editor

vegorovinno:


NathanSweet:
This is not a bad idea, but needs to be done for all runtimes. For JSON we could try to parse, then if parsing fails we could wrap the parse error with an "versions don't match, JSON version: x, runtime version: y" message.

badlogic:
Closed in favor of this issue https://github.com/EsotericSoftware/spine-runtimes/issues/1497

PR #1447, opened 2019-08-13
Make JsonRollback results support 3.6 runtime

taigacon:
Make JsonRollback generated results support 3.6 runtime. (3.6.53)


NathanSweet:
Likely 3.7 JSON will work in 3.6. If you can post 3.7 JSON that doesn't work in 3.6, we'll reopen this issue.

Note that using the JsonRollback tool is not a workflow that you should rely on.

taigacon:
Sorry for the late reply.
The 3.7 json can not work perfectly in 3.6, because of the lack of "x" and "y" field in the "scale“ group.
It seems that editor/runtime will treat the lack as default value.
However, the default values are 1.0/1.0 in 3.7, and 0.0/0.0 in 3.6.

taigacon:
After the translate, the animation of the project skeleton_3.6.53 is broken due to the scale set to 0/0.

[project.zip](https://github.com/EsotericSoftware/spine-runtimes/files/3521739/project.zip)

NathanSweet:
A recent commit has fixed this, please try now (the Skeleton Viewer has been updated) or use this 3.7 JSON:
http://n4te.com/x/416-skeleton_3.7.json

taigacon:
Sure, thanks a lot.

PR #1450, opened 2019-08-14
[unity] BoneFollower support as a multi-level child

kvfreedom:


HaraldCsaszar:
BoneFollower already supports multi-level children, else branch is responsible if BoneFollower is not an immediate child, see https://github.com/EsotericSoftware/spine-runtimes/blob/3.8/spine-unity/Assets/Spine/Runtime/spine-unity/Components/Following/BoneFollower.cs#L159.

Implementation of pull request leads to incorrect result if intermediate parents have non-identity transforms, assigning local transformation value (result is off by parent transform).

PR #1461, opened 2019-08-21
Fixes the situation where the timeline, which is not attachment type,…

lolisia:
… mixes with Setup

Fixes the situation where the timeline, which is not attachment type, mixes with Setup

### DIFF
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -299,14 +299,18 @@ namespace Spine {
 					float alpha;
 					switch (timelineMode[i] & AnimationState.NotLast - 1) {
 						case AnimationState.Subsequent:
-							if (!attachments && timeline is AttachmentTimeline) {
-								if ((timelineMode[i] & AnimationState.NotLast) == AnimationState.NotLast) continue;
-								blend = MixBlend.Setup;
+							{
+								var subsequentBlend = blend;
+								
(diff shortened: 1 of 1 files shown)


NathanSweet:
AnimationState is very complex and there are many scenarios it needs to support. it's very easy to break other scenarios when making a change. Can you please show the scenario that doesn't work right which this PR fixes?

lolisia:
When attachment timeline mixing with Setup Pose, '**blend**' variable is written to MixBlend.Setup, and subsequent **AnimationState.Subsequent** timelines are all mixed with MixBlend.Setup.

My guess is that if it's not an AttachmentTimeline or a DrawOrderTimeline, you'll need to use the value of the '**blend**' variable you determined above on the Timeline Loop, **not MixBlend.Setup**.

lolisia:
Please note that due to company rules, I cannot upload resources under development.

It is thought that this occurs when A → B → C SetAnimation has a longer Mix Duration of B → C than the playback time of A → B section.

When B → C Mix, Timeline such as Rotate, Translate, etc. was expected not to mix with Setup Pose. However, when the Mix Duration is long after migration with Spine 3.8 Runtime, it is operating to mix with Setup Pose in most sections.

NathanSweet:
Thanks for the explanation. I believe you are right -- an attachment timeline should not change the blend used for other timelines. We've made the changes for all runtimes. Cheers!

PR #1476, opened 2019-09-03
compare using pre-calculated hashCode

lolisia:
compare using pre-calculated hashCode


badlogic:
I'm afraid this implementation would violate the `Equals` contract in that two different `SkinEntry` instances could have the same hash code.

PR #1479, opened 2019-09-04
Remove unnecessary STL includes

stienb:
This pull request removes some uses of the STL that are not necessary.

Most of them were just includes left there in error.

`std::swap` was used for real in Vector.h.
It is replaced in a simple way, but needs the c++ version to be pushed-up to `c++11`, though.

Removing these STL includes translates to smaller compilation times for large projects, since the size of the preprocessed source is much smaller.

### DIFF
--- a/spine-cpp/spine-cpp/include/spine/RTTI.h
+++ b/spine-cpp/spine-cpp/include/spine/RTTI.h
@@ -32,16 +32,27 @@
 
 #include <spine/SpineObject.h>
 
+#ifdef SPINE_UE4
 #include <string>
+#else
+#include <spine/SpineString.h>
+#endif
 
 namespace spine {
+
+#ifdef SPINE_UE4
+typedef std::string StringImpl;
+#else
+typedef String StringImpl;
+#endif
+
 class SP_API RTTI : public SpineObject {
 public:
-	explicit RTTI(const std::string &className);
+	explicit RTTI(const StringImpl &className);
 
-	RTTI(const std::string &className, const RTTI &baseRTTI);
+	RTTI(const StringImpl &className, const RTTI &baseRTTI);
 
-	const std::string &getClassName() const;
+	const StringImpl &getClassName() const;
 
 	bool isExactly(const RTTI &rtti) const;
 
@@ -53,7 +64,7 @@ private:
 
 	RTTI &operator=(const RTTI &obj);
 
-	const std::string _className;
+	const StringImpl _className;
 	const RTTI *_pBaseRTTI;
 };
 }
--- a/spine-cpp/spine-cpp/src/spine/RTTI.cpp
+++ b/spine-cpp/spine-cpp/src/spine/RTTI.cpp
@@ -32,17 +32,16 @@
 #endif
 
 #include <spine/RTTI.h>
-#include <spine/SpineString.h>
 
 using namespace spine;
 
-RTTI::RTTI(const std::string &className) : _className(className), _pBaseRTTI(NULL) {
+RTTI::RTTI(const StringImpl &className) : _className(className), _pBaseRTTI(NULL) {
 }
 
-RTTI::RTTI(const std::string &className, const RTTI &baseRTTI) : _className(className), _pBaseRTTI(&baseRTTI) {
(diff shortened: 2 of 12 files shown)


stienb:
Bump.

spine-cpp/spine-cpp/include/spine/RTTI.h
```diff
@@ -31,17 +31,16 @@
 #define Spine_RTTI_h
 
 #include <spine/SpineObject.h>
-
-#include <string>
+#include <spine/SpineString.h>
```
badlogic, line 38: This needs to be `std::string` due to initialization order in some engines (UE4)

spine-cpp/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ if(MSVC)
 	set (CMAKE_C_FLAGS "${CMAKE_CXX_FLAGS}")
 else()
 	set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -pedantic -std=c89")
-	set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wnon-virtual-dtor -pedantic -std=c++03 -fno-exceptions -fno-rtti")
+	set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wnon-virtual-dtor -pedantic -std=c++11 -fno-exceptions -fno-rtti")
```
badlogic: We need to enforce C++03 due to compiler toolchains a few of our customers use.

spine-cpp/spine-cpp/src/spine/RTTI.cpp
```diff
@@ -32,17 +32,16 @@
 #endif
 
 #include <spine/RTTI.h>
-#include <spine/SpineString.h>
 
 using namespace spine;
 
-RTTI::RTTI(const std::string &className) : _className(className), _pBaseRTTI(NULL) {
```
badlogic, line 39: Same as above. These need to stay std::string.

badlogic:
Please don't bump. Everyone following this project will receive an email.

stienb:
Thanks for the feedback, I appreciate.
I took your comments into account, sticked to c++03, and included std::string only for spine-ue4.

Removing STL includes when they're not needed can improve game compile times very significantly. Of course for people that already use the STL pretty much everywhere in their codebase this is irrelevant...

Since you guys already did the hard work with spine::Vector and spine::String, it would really be too bad to keep the last bits of STL in there!

To illustrate this, here's a table showing the number of lines of code that are actually compiled in spine-cpp after preprocessing.

|     |compiled lines of code<br>without STL includes |compiled lines of code<br>with STL includes |
|-----|--------------------------------------|-----------------------------------|
|spine-cpp|                               360,867|                          1,466,818|

And that's just the libspine-cpp library.
The impact on a large codebase which heavily includes spine headers can be even greater.

Hope this helps!

Below are the small scripts used to generate this:
```count_preprocessed.sh```
```
#!/usr/bin/env bash

if [ "$#" -ne 1 ]; then
    echo "usage: count_preprocessed.sh <cpp_file>"
    echo
    echo "This will count the number of lines of code in cpp_file after preprocessing."
    echo
    echo "example: count_preprocessed.sh spine-cpp/src/spine/Skeleton.cpp"
    exit 1;
fi

cpp_file="$1"
preprocessed_file="${cpp_file%.*}.i"
file_prefix="CMakeFiles/spine-cpp.dir/"
make $preprocessed_file && wc -l ${file_prefix}$preprocessed_file |  awk '{printf "%s\t%s\n", $1, "'$cpp_file'"}'
rm CMakeFiles/spine-cpp.dir/$preprocessed_file
```

and:
```count_all.sh```
```
#!/usr/bin/env bash
for file in `find spine-cpp/src/spine -type f -name "*.cpp"`; do ./count_preprocessed.sh $file; done | grep -v Preprocessing | sort -k2
```

badlogic:
Getting rid of STL is definitely a big win. I went through the code base and removed all STL includes and resolved the RTTI string conundrum by using good old static string literals. Thanks for your help!

PR #1484, opened 2019-09-06
Fix cocos2dx project and use of premultiplied alpha

sp-stefano-lanza:
This PR contains a couple of fixes:
1. It fixes the spine-cocos2d-x.vcxprox Visual Studio project
2. In cocos2dx ```SkeletonRenderer```, take into account whether a texture is encoded with premultiplied alpha when setting the blending mode.


badlogic:
Cheers!

PR #1500, opened 2019-09-20
Add support for webgl2 context

MikalDev:
Add support for webgl2 context, needed for integration with rendering in Construct 3 gl context (C3 defaults to a webgl2 context if available.) C3 users are creating a C3 plugin using Spine-TS, WIP, the basics working.


badlogic:
I haven't dabbled in WebGL 2 much yet, but I'd assume this is only one of a few changes to have full WebGL 2 support, correct? Thinking of VBO vs VAO and some changes to vertex shaders (attribute specifiers etc.).

I will add this simple fix manually myself and see about the rest.

MikalDev:
I think just to get the render working in a webgl2 context this may be the only change since  webgl2 is _almost_ 100% backward compatible to webgl1 and in general, this is all it took to get SceneRenderer.drawSkelton to work (though on your suggestion, I am going to look at just using SkeletonRenderer instead going forward.)
[webgl2 back compat](https://webgl2fundamentals.org/webgl/lessons/webgl1-to-webgl2.html)

PR #1533, opened 2019-10-28
Set image source after crossOrigin and event handlers

matzematthaey:
This would allow proper crossOrigin image loading. Otherwise the browser would reject loading images due to security concerns.


badlogic:
Cheers! Sorry I overlooked this.

PR #1540, opened 2019-11-01
[libgdx] Fix skeletons not being rendered properly to FBOs due to blend functions.

rafaskb:
When drawing skeletons to a FBO with BlendMode.normal, transparency is created when drawing a image with transparency on top of an opaque one. Depending on the background this artificial transparency looks like artifacts.

This happens because SkeletonRenderer uses a combined blend function. Switching to separate blend functions and adjusting BlendMode to use GL_ONE instead of GL_ONE_MINUS_SRC_ALPHA on dstAlpha fixes the issue.

blending_bug

image

The same issue is also present throughout libGDX, which results in FBOs never working properly out of the box.

### DIFF
--- a/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/BlendMode.java
+++ b/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/BlendMode.java
@@ -33,26 +33,38 @@ import com.badlogic.gdx.graphics.GL20;
 
 /** Determines how images are blended with existing pixels when drawn. */
 public enum BlendMode {
-	normal(GL20.GL_SRC_ALPHA, GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_ALPHA), //
-	additive(GL20.GL_SRC_ALPHA, GL20.GL_ONE, GL20.GL_ONE), //
-	multiply(GL20.GL_DST_COLOR, GL20.GL_DST_COLOR, GL20.GL_ONE_MINUS_SRC_ALPHA), //
-	screen(GL20.GL_ONE, GL20.GL_ONE, GL20.GL_ONE_MINUS_SRC_COLOR), //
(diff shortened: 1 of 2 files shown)


NathanSweet:
Note only non-PMA rendering needs the separate blend functions when writing to an FBO, and then when rendering the FBO PMA blending is needed (ONE, ONE_MINUS_SRC_ALPHA). When using PMA images, rendering to the FBO can use PMA blending.

NathanSweet:
There appears to be an error in the PR, which has non-PMA `normal` using:
```
color src, dst: SRC_ALPHA, ONE_MINUS_SRC_ALPHA
alpha src, dst: GL_SRC_ALPHA, GL_ONE
```
We want:
```
color src, dst: SRC_ALPHA, ONE_MINUS_SRC_ALPHA
alpha src, dst: ONE, ONE_MINUS_SRC_ALPHA
```

Simple libgdx test app and image:
http://n4te.com/x/1388-cUm8.txt
http://n4te.com/x/1389-circle.png

Code that can be pasted into SkeletonViewer to have it first render to an FBO:
```
FrameBuffer fbo = new FrameBuffer(Format.RGBA8888, 120, 120, false);
fbo.begin();
Gdx.gl.glClearColor(0, 0, 0, 0);
Gdx.gl.glClear(GL_COLOR_BUFFER_BIT);
batch.getProjectionMatrix().setToOrtho2D(0, fbo.getHeight(), fbo.getWidth(), -fbo.getHeight());
batch.begin();
renderer.draw(batch, skeleton);
batch.end();
fbo.end();
fbo.getColorBufferTexture().setFilter(TextureFilter.Nearest, TextureFilter.Nearest);
batch.getProjectionMatrix().set(camera.combined);
batch.setBlendFunction(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
batch.begin();
batch.draw(fbo.getColorBufferTexture(), 0, 0);
batch.end();
fbo.dispose();
```
Spine project and export for Skeleton Viewer:
http://n4te.com/x/1390-yt.zip

NathanSweet:
Fixed in 4.0.

PR #1244, opened 2019-01-08
[cocos2d-x] adpater to new renderer in cocos2dx v4

PatriceJiang:
~~Cocos2d-X new renderer for Metal is in developing.  https://github.com/minggo/cocos2d-x/tree/metal-support~~

~~This PR works only on the new renderer, and we may need a new branch, like `3.7-cocos2d-x-v4`.~~

Current metal support branch is https://github.com/cocos2d/cocos2d-x/tree/v4

PR #1572, opened 2019-12-13
Updated CMake scripts and cocos2d-x v3 and v4 specific file paths.

rh101:
@badlogic As discussed in https://github.com/EsotericSoftware/spine-runtimes/pull/1244#issuecomment-565471797 

This setup works, so there should be no need to have separate include statements for v3 and v4.  I've left the `examples` and `examples-v4` in there because, as you pointed out, there are minor differences in certain platform specific files between Cocos2d-x v3 and v4, and I wasn't quite sure if the newer v4 stuff would work with v3.

Other than that, some minor changes to the spine-cpp and spine-cocos2dx CMakeList.txt files that may make it easier to include them into an application.


badlogic:
Finally had time to look into this. I've taken some of your CMake changes. While the preprocessor juggling to include the correctly versioned batchers looks nasty on first sight, I believe for our end users it's the easier approach. Most of them will simply copy the entire ``spine-cocos2dx/src` and `spine-cpp/spine-cpp/src` directories into their source tree and call it a day instead of fiddling with CMake. While your solution is exceedingly more clean, I fear that it would lead to more support requests. Thanks for the PR though, really appreciate your help and I was able to clean up some parts of the existing CMake files!

PR #1584, opened 2019-12-31
use std::function when compiler support c++11

codetypess:
Since c++11, std::function is supported, so maybe better use '__cplusplus >= 201103L' instead of define 'SPINE_USE_STD_FUNCTION' macro


badlogic:
We do not want to force the use of `std::function` when compiling with C++11 as that will break existing applications.

PR #1587, opened 2020-01-02
fix: [ts] Skin.attachAll: empty skin placeholders wont be reattached after a skin change

modjke:
Exactly what title says
Steps to reproduce:
Spine with at least two skins
Skin A: skin placeholder with no attachement
Skin B: above skin placeholder, but with attachement
```
setSkinByName('Skin A')
setSkinByName('Skin B')
```
Skin attachement will not be attached and will not be visible

### DIFF
--- a/spine-ts/core/src/Skin.ts
+++ b/spine-ts/core/src/Skin.ts
@@ -194,6 +194,12 @@ module spine {
 							break;
 						}
 					}
+				} else {
+					let attachmentName = slot.data.attachmentName; 
+					if (attachmentName != null) {
+						let attachment = this.getAttachment(slotIndex, attachmentName);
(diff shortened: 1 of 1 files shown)


modjke:
Umm, is this being reviewed or something? I have at least two projects affected by this right now.

badlogic:
This is a non-trivial change that will affect all users for the runtime. We've been discussing this internally and will maybe introduce the change in the 4.0 release later this year (currently 3.9-beta branch). For 3.8, we can't apply this change for API stability reasons. I suggest you apply the change locally for the time being.

badlogic:
We are in the process of updating the runtimes for the 4.0 release, which will cover this PR. Closing this out.

PR #1619, opened 2020-02-18
fix: can not clear track when it's end time is reached

ghost:


badlogic:
Could you elaborate on what the problem is? The spine-cpp runtime follows the reference implementation, and this would mean a change in behaviour.

badlogic:
This is by design, see the reference implementation.

PR #1624, opened 2020-02-25
[unity] SkeletonRenderer not generating mesh

lolisia:
Fixed a situation where skeleton renderer used in prefab / scene does not generate mesh


HaraldCsaszar:
@lolisia Thanks for reporting. Unfortunately, while your commit might solve one problem, it causes a problem when dragging a prefab into a scene, the instance in the scene will then not load the mesh. Simply inverting the logical-not is not a solution here.

Please provide a reproduction project where your problem described above can be reproduced, we will then be happy to reopen the ticket and fix the issue.

PR #1627, opened 2020-02-26
Fixing memory leaks issue #1623 [runtimes]

### LINKED ISSUE #1623: [cpp] Memory leak in SkeletonBounds
Polygons which allocated and stored in _polygonPool vector gets never deleted causing memory leaks.

DevJ0049:
Fix #1623 [runtimes]


badlogic:
Fixed as per issue.

PR #1662, opened 2020-04-22
[spine-c] Fixed uninitialized variable timeline

kerrytazi:
c runtime crashes because of uninitialized variable timeline.


JiriAtanasovsky:
There seems to be same bug in LUA version.

AnimationState.lua on line 383 is missing timeline variable to be passed.

badlogic:
Sorry, overlooked this. Fixed!

PR #1663, opened 2020-04-22
[spine-c] Fixed path constraint json name

kerrytazi:
spine-c uses wrong json key for path constraints.


badlogic:
Sorry, this has already been fixed.

kerrytazi:
I don't see commits in 3.8 or 3.9-beta

PR #1680, opened 2020-05-18
[TS] Fixed that items in a pool were replaced when calling `freeAll`

smaren:
A call to free all would replace items currently in the pool. This change calls `free` instead which adds the items to the pool instead.


badlogic:
Cheers, I've fixed this up manually, as the use of `Pool.free()` already does the reset call.

PR #1689, opened 2020-06-03
SkeletonRenderer::LateUpdate() does not work when the target is an in…

lolisia:
…stance of Prefab.

SkeletonRenderer::LateUpdate() does not work when the target is an instance of Prefab.


HaraldCsaszar:
Thanks for reporting!

PR #1690, opened 2020-06-03
[spine cpp] Add object release hook

PatriceJiang:
In our application, a `SpineObject` can be connected to a JavaScript Object, it need to be unconnected when destroyed. So I need get notification before destroy.


badlogic:
Great addition, thanks!

PR #1691, opened 2020-06-03
add function AnimationState:getLast (trackEntry)

JiriAtanasovsky:
I found usefull to have function that returns trackEntry for last animation queued.

I use this to get duration of last animation queued to adjust delay to be from END of the last animation instead of its START to get its duration.


badlogic:
Cheers!

PR #1697, opened 2020-06-09
[lua] Fixed missing call to set attachment method

iwiniwin:
Fixed missing call to set attachment method, resulting in abnormal animation effect


badlogic:
Thanks! Fixed in the 3.8 and 4.0-beta branches.

PR #1698, opened 2020-06-11
SkeletonInput::ReadFloat() Optimization

lolisia:
I want to optimize Stream.ReadByte() repeat calls by calling a lot of ReadFloat() when configuring large-scale animation/Deform information.


HaraldCsaszar:
Thanks for the hint @lolisia!

PR #1706, opened 2020-06-18
[spine-c] Fixed floating point exception - division by zero

kerrytazi:
_isNan throws exception when FP exceptions (FE_DIVBYZERO) are enabled.


badlogic:
Thanks!

PR #1709, opened 2020-06-23
Optimize animation constructor performance

lolisia:
Optimization to avoid unnecessary GC occurrence by calling hashset repeatedly SetCapacity()


HaraldCsaszar:
Thanks for the hint. While we don't see a dramatic reduction of GC allocation, the effect may be more noticable under some circumstances. We chose an implementation without using LINQ which creates even fewer allocations.

PR #1710, opened 2020-06-23
Replace deprecated API

HustLion:
`MALLOC` not found in cocos2dx 3.10


badlogic:

PR #1711, opened 2020-06-23
fixed duplicate namespace specifier

HustLion:
This duplication causes compile error.


badlogic:
Thanks, fixed in the 3.8 and 4.0-beta branches. Could you tell me on which compiler this errored for you? I couldn't reproduce this with neither Clang nor MSVC.

HustLion:
> Thanks, fixed in the 3.8 and 4.0-beta branches. Could you tell me on which compiler this errored for you? I couldn't reproduce this with neither Clang nor MSVC.

I'm using VS2013, which is a legacy version since our team is using the old cocos2dx-3.10 engine.

PR #1712, opened 2020-06-23
Optimization of reuse of empty deform buffer

lolisia:
If there is no vertex information in the deform timeline, and weights are used, a new float[deformLength] with an empty value is used and the value is not updated later. The goal is to reuse it and optimize it.

Below is a screenshot of profiling when initializing SkeletonData for one character in the current project.
In the case of a lot of Animation information, we could reduce GC of 10MB by reusing the buffer. (It's still too big, so we need a lot of optimization)

before
after

### DIFF
--- a/spine-csharp/src/SkeletonBinary.cs
+++ b/spine-csharp/src/SkeletonBinary.cs
@@ -64,6 +64,8 @@ namespace Spine {
 		private AttachmentLoader attachmentLoader;
 		private List<SkeletonJson.LinkedMesh> linkedMeshes = new List<SkeletonJson.LinkedMesh>();
 
+		private static readonly Dictionary<int, float[]> emptyFloatBuffer = new Dictionary<int, float[]>();
+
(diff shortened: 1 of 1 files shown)


HaraldCsaszar:
Thanks for your efforts. Unfortunately the proposed solution does not clear the dictionary at any time, while clearing it (e.g. at the end of the `Animation` ctor) would then lead to a memory leak, potentially triggering GC.

In general, these `float[]` allocations should not be problematic unless you unload your skeletons frequently mid-game instead of at level-changes.

HaraldCsaszar:
May I ask what your use case scenario is that is causing problems with the memory allocated by the skeleton loader? Are you actually unloading skeletons frequently? If so - why?

Apart from that, 435 animations at a single skeleton is a lot! Are you sure that you need all these animations and have to use vertex deforms and cannot go with weighted bone rotation/scale animation instead?

PR #1732, opened 2020-08-05
Add missing const-qualifiers

halx99:


halx99:
it's important for passing EventData as const reference, it's base of c++ concept, I want to tell you wen I  integrate spine-runtime-3.8 to latest cocos2d-x-v4, can't be compile if I want enable luabindings and the official cocos2d-x-4.0 disable spine 3.8 luabinding on it's cmake script by default.
That is why I post this PR, without this changes, user can't use spine-runtime-3.8 for cocos2dx lua bindings API.
So don't forget sync this PR changes

badlogic:
Sorry, I didn't mean to close the PR! I'll have to do a const correctness pass eventually.

halx99:
it's ok, just do our best

PR #1733, opened 2020-08-05
cocos2d-x v4 no js binding

halx99:


badlogic:

PR #1734, opened 2020-08-05
Update spine-cocos2dx.cpp, no need to call fullPathForFileName

halx99:


badlogic:

PR #1736, opened 2020-08-10
[spine-cpp] remove usage of nullptr

FredGithub:
spine-cpp does not compile on linux because `nullptr` is not defined for -std=c++03. This pull request simply replaces it with `NULL`


badlogic:
Cheers, this has been fixed in 3.8 and 4.0-beta.

PR #1748, opened 2020-08-27
Fix problems when play animation in reverse

jingqi:
Problems are: non-looping reverse play will stop at T-Pose, not the first frame as expected; looping reverse play won't loop and stopped at first round


badlogic:
`AnimationState` does not support playing animations in reverse if an animation includes specific types of timelines. Your fix will work for some trivial animations, but fall apart for others. We suggest to create a duplicate animation Spine and reversing the keys instead of trying to get this right at runtime.

PR #1761, opened 2020-09-15
[ts] Enable SharedAssetManager to work in a WebWorker

MikalDev:
Adding Web Worker support for spine-ts in the SharedAssetManager class, which is used in the Construct 3 Spine plugin. This enables us to use spine-ts and the Spine addon with Construct 3 in Web Worker mode (which can reduce FPS jank in a web browser, decoupling from the main browser thread.)

I did not do the changes in the AssetManger, as the C3 Spine plugin does not use that class, so I cannot test the changes (the changes also look more extensive there.)

Tested with the Construct 3 plugin in web worker mode and in non web worker mode on Chrome, Safari, iOS Safari.

The main changes for web worker: use fetch() instead of img.src, use ImageBitMap instead of HTMLImageElement, enable OffscreenCanvas instead of HTMLCanvasElement.

Construct 3 web worker description here:
https://www.construct.net/en/blogs/ashleys-blog-2/performance-isolation-html5-1512


MikalDev:
Some of the conflicts to be resolved for spine-ts/core/src/SharedAssetManager.ts don't look correct. The merge diffs look correct though. I can redo this if needed.

MikalDev:
Going to look at cleaning up some of the code to get some cleaner resolve outcomes if possible. Also make sure fork has pulled all the commits. Closing this PR.

PR #1762, opened 2020-09-15
[ts] Enable SharedAssetManager to work in a WebWorker

MikalDev:
Adding Web Worker support for spine-ts in the SharedAssetManager class, which is used in the Construct 3 Spine plugin. This enables us to use spine-ts and the Spine addon with Construct 3 in Web Worker mode (which can reduce FPS jank in a web browser, decoupling from the main browser thread.)

I did not do the changes in the AssetManger, as the C3 Spine plugin does not use that class, so I cannot test the changes (the changes also look more extensive there.)

Tested with the Construct 3 plugin in web worker mode and in non web worker mode on Chrome, Safari, iOS Safari.

The main changes for web worker: use fetch() instead of img.src, use ImageBitMap instead of HTMLImageElement, enable OffscreenCanvas instead of HTMLCanvasElement.

Construct 3 web worker description here:
https://www.construct.net/en/blogs/ashleys-blog-2/performance-isolation-html5-1512


MikalDev:
Hi, checking if there is any feedback on this PR. I signed the CLA.

badlogic:
Sorry, haven't had time to get to it yet. Will try to do so this week!

badlogic:
Finally got around to this. Looking good! I had to make minor changes to the canvas backend which should be OK. Thank you for the PR!

MikalDev:
Thanks for accepting the PR (and doing the changes for the canvas backend.) Now I can keep the C3 spine plugin in sync w/ latest releases (and move to 4.0 when the time is right.)

NathanSweet:
@MikalDev Heads up, we've removed `SharedAssetManager` since now `AssetManager` has similar functionality: it avoids downloading the same assets multiple times if you pass the same `Downloader` instance to each of your `AssetManager` constructors. I ported the webworker code to AssetManager, though it's untested.

PR #1763, opened 2020-09-16
[WIP][cocos2dx-v4] Re-enable support for a custom SpineAnimation shader 

rh101:
These changes re-enable support for a custom shader applied to SpineAnimation when using SkeletonBatch with Cocos2d-x v4. This won't have any affect when using SkeletonTwoColorBatch, since that is already applying its own shader for the two color effect.

NOTE: When a custom shader is applied, it breaks batching, because of this code in Cocos2d-x, `CCTrianglesCommand.cpp`:
```
void TrianglesCommand::init(float globalOrder, Texture2D* texture, const BlendFunc& blendType, const Triangles& triangles, const Mat4& mv, uint32_t flags)
{
...
        
        //since it would be too expensive to check the uniforms, simplify enable batching for built-in program.
        if(_programType == backend::ProgramType::CUSTOM_PROGRAM)
            setSkipBatching(true);
        
...
}
```

Not sure what to do about that.


rh101:
To fix the batching with custom shaders, changes to cocos2d-x v4 engine code had to be made, and are now in a pull request [https://github.com/cocos2d/cocos2d-x/pull/20584](https://github.com/cocos2d/cocos2d-x/pull/20584)

With the changes to the cocos2d-x engine code in that pull request, multiple spine animations will use a single draw call if the custom shader has the same uniform values.

badlogic:
This looks great! I'll have to validate the PR, but from a quick look I think we can merge this as is. Thanks!

rh101:
It does work as it is.  The only issue is that the attachments on the Spine animation, one that is using a custom fragment shader, do not get batched into one draw call, even if they are using the same texture, but that has nothing to do with Spine or the changes in this pull request; it's a Cocos2d-x v4 issue, and if and when the Cocos2d-x devs fix it (such as the Cocos2d-x pull request I linked above), no modifications should be required to Spine in order to batch the draw calls.

I'm using it at the moment with the modified Cocos2d-x v4, and every sprite animation with a unique fragment shader uses only a single draw call per shader (which is correct).

badlogic:
Hm, IIRC, when I tested the v4 spine-cocos2dx runtime, the old way via `TriangleCommand` was batching correctly even with the custom shader. I assume this no longer works? I'll find out when I merge this.

badlogic:
@halx99 @rh101 i'm trying to understand how your two PRs (this one and #1787) interact with each other. It looks like @halx99 's PR supersedes this PR, is this correct?

badlogic:
As per @halx99 we are going with his PR, as it avoids a few performance issues.

rh101:
I'm not sure about the implementation by @halx99 since I wasn't involved in creating it.  If that implementation is more suitable, then go with it that one.  The code by @halx99 seems similar, with a few additions like overriding the SkeletonTwoColorBatch shader.  We just need the custom shader support re-enabled in the cocos2d-x v4 runtime, and how it's done is entirely up to you.

Regarding the UniformLocation of both the MVP and texture, will it always be the same for all program states?  If so, then the `CustomTrianglesCommand` I added in my version wouldn't be required.

```
	_locMVP = programState->getUniformLocation("u_MVPMatrix");
	_locTexture = programState->getUniformLocation("u_texture");
```

badlogic:
I'd assume the uniform location might change on some platforms, i.e. if the OpenGL context is lost on older Android versions.

Issue #1784, opened 2020-10-15
[unity] The SkeletonMecanim does not handle negative normalizedTime


stefanbursuc:
The SkeletonMecanim does not properly handle animation state's normalizedTime when it goes negative.

The animation freezes on the first animation frame.

This behavior can be viewed on this video: https://streamable.com/gg3926.
And can be tested on the attached test project on this forum thread:
http://esotericsoftware.com/forum/Bug-The-SkeletonMecanim-does-not-handle-negative-times-14744

stefanbursuc:
I have proposed a fix with the pull request #1785.

HaraldCsaszar:
Fixed in this issue ticket:
https://github.com/EsotericSoftware/spine-runtimes/issues/1790

stefanbursuc:
Sorry for the late reply. 
The fix for this bug added a regression.
Please test again the provided scene from the forum thread mentioned above.
Change the Animator's "New State" Speed value (not the Multiplier) to a negative value (i.e. -1).
The animation should go backwards with a positive multiplier (change-able from the Game View) and forwards for negative multiplier.

stefanbursuc:
I have created a follow-up issue: #1810

HaraldCsaszar:
Thanks for reporting @stefanbursuc !

PR #1785, opened 2020-10-15
Fix SkeletonMecanim handling of negative normalizedTime value.

stefanbursuc:
This is a fix for ticket: #1784.

The AnimationTime needs to also handle negative values for the current clip's **normalizedTime**.
This can happen if someone uses a negative value for the animation's speed multiplier in order to
play the animation in reverse.

Can be rewritten in some other form.


HaraldCsaszar:
Fixed in this issue ticket:
https://github.com/EsotericSoftware/spine-runtimes/issues/1790

PR #1786, opened 2020-10-15
Update main.cpp

dsvi:
Otherwise "use after free" happens. When global `RTTI::~RTTI` calls 

SpineObject::~SpineObject() {
	SpineExtension::beforeFree(this);
}

But `dbgExtension` is dead at that time already. So call to `getInstance()` in `beforeFree()` returns pointer to garbage.

### DIFF
--- a/spine-sfml/cpp/example/main.cpp
+++ b/spine-sfml/cpp/example/main.cpp
@@ -612,9 +612,8 @@ void test (SkeletonData* skeletonData, Atlas* atlas) {
 
 int main () {
 	DebugExtension dbgExtension(SpineExtension::getInstance());
-	SpineExtension::setInstance(&dbgExtension);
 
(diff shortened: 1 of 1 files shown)


badlogic:
How can `dbgExtension` be dead when the call to `SpineExtension::setInstance()` is made? It will be RAIIed once `main` is done.

badlogic:
Ah, I see now, SwirlEffect is static and destructed after the debug extension. Your fix isn't entirely correct though, as the debug extension would never clean up this way nor be used. I'll fix it!

dsvi:
> How can dbgExtension be dead when the call to SpineExtension::setInstance() is made? It will be RAIIed once main is done.

Quite simply compiler doesn't care about what call you have made. Once object got out of scope (e.g. after the closing } ), it gets destroyed.

PR #1795, opened 2020-10-22
Add compiler macro SPINE_DATA_FORMAT_VERSION check to compatible with…

halx99:
… 3.7 data load runtime

### DIFF
--- a/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
+++ b/spine-cocos2dx/src/spine/SkeletonRenderer.cpp
@@ -197,7 +197,11 @@ void SkeletonRenderer::initWithJsonFile (const std::string& skeletonDataFile, At
 }
 
 void SkeletonRenderer::initWithJsonFile (const std::string& skeletonDataFile, const std::string& atlasFile, float scale) {
+#if !defined(SPINE_DATA_FORMAT_VERSION) || SPINE_DATA_FORMAT_VERSION >= 0x0308
 	_atlas = new (__FILE__, __LINE__) Atlas(atlasFile.c_str(), &textureLoader, true);
+#else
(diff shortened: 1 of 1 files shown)


halx99:
this pr can allow user easy switch between 3.7 and 3.8 data searlization runtime to work with cocos2dx 4.0

badlogic:
Sorry, a specific version x.y of the Spine Runtimes is only supposed to support exports from the corresponding Spine Editor version  x.y. It's not just the data format that might change between versions, but also behaviour and how the data is interpreted. We will not support multiple Spine Editor export versions in a single Spine Runtime version.

halx99:
> Sorry, a specific version x.y of the Spine Runtimes is only supposed to support exports from the corresponding Spine Editor version x.y. It's not just the data format that might change between versions, but also behaviour and how the data is interpreted. We will not support multiple Spine Editor export versions in a single Spine Runtime version.

So if user want use spine-3.7, they have no chance to use cocos2dx v4?

badlogic:
3.7 and 3.8 are similar enough that a fix like yours might work. However,
that is likely not true for other version pairs. If we start doing it for
3.7 and 3.8, users expect us to do it for all version pairs, which is not
possible and complicates the runtimes further.

If you think cocos2d-x v4 support for exports from Spine 3.7 is important,
we can talk about adding that support in the 3.7 spine-runtimes branch. Buy
supporting multiple export versions in a specific runtimes version will not
happen I'm afraid.

halx99:
thanks for your explain, I got it

halx99:
Agree, support cocos2dx v4 on 3.7 branch is better

halx99:
And I have finish both 3.6, 3.7 and 3.8 works on adxeproject (which is a fork of cocos2d-x-4.0):
- spine 3.6~3.8 for adxe: https://github.com/adxeproject/spine-adxe

PR #1801, opened 2020-10-27
Cocos2dx v4 custom shader fix

rh101:
This address the problem described [here](https://github.com/EsotericSoftware/spine-runtimes/pull/1787#issuecomment-716964507)


badlogic:
Thanks. I need to add a multi-page atlas export to the tests for this kind of error to be caught. Sorry for the hiccup!

PR #1808, opened 2020-11-10
[ue4] Fixed shadow declaration warnings

nprudnikov:
I suggest removing variables from the global scope. When building spine + ue4, you can get warnings:
```
warning C4459: declaration of 'First' hides global declaration
warning C4459: declaration of 'Current' hides global declaration
```

PS. The best practice would be to wrap these variables in an enum, but that would require a little more changes. Let me know if you need it.


HaraldCsaszar:
Thanks for reporting! It has just been fixed on the 3.8 and 4.0-beta runtimes via the above commit, implemented slightly different compared to your pull request to keep differences between runtime implementations small.

PR #1846, opened 2021-02-19
[cpp] Allow ownership of AttachmentLoader to be passed to SkeletonBinary and SkeletonJson

rh101:
There are situations where a custom `AttachmentLoader` is required, yet there is no way to pass ownership to the instance of this class to `spine::SkeletonBinary` and `spine::SkeletonJson`.  The only way `spine::SkeletonBinary` and `spine::SkeletonJson` can have ownership is if they internally create their own instance of `AttachmentLoader`, which is specifically of type `AtlasAttachmentLoader`.

The API has been slightly modified to allow ownership of the `AttachmentLoader` to be passed to `spine::SkeletonBinary` and `spine::SkeletonJson`, without breaking any current implementations using it.


badlogic:
This has been added to spine-cpp a while ago!

rh101:
> 
> 
> This has been added to spine-cpp a while ago!

@badlogic I'm not sure what you mean.  Can you please explain how, because I am not seeing how it's possible to pass ownership of the custom attachment loader to those 2 classes.  `_ownsLoader` is private, and there is no way to set it to true.  The only time it is set to true is if this constructor is called with the default `AtlasAttachmentLoader`:

```
SkeletonJson::SkeletonJson(Atlas *atlas) : _attachmentLoader(new(__FILE__, __LINE__) AtlasAttachmentLoader(atlas)),
	_scale(1), _ownsLoader(true)
{}
```
Same issue with `SkeletonBinary`.

That means we cannot pass our own attachment loader to it and also pass ownership of that attachment loader.

badlogic:
Oops, my bad, wasn't looking at the private qualifier. Merged, cheers!

rh101:
Thank you!

PR #1851, opened 2021-03-01
[TS] Support applying premultiply on loading images to fix blending

sp-jesus-diaz:
### WHY: 
After researching how the blending works, we found out that premultiplying alpha is **required** to apply **multiply**/**additive** blending options and see Slots correctly. 
Because re-generating and re-deploying assets changing **premultiply** is not always possible or trivial to do, it makes sense to apply "premultiply" when loading Images on browser.  

### HOW:
- Add optional `bitmapOptions` parameter to `AssetManager.loadTexture` to control how bitmaps are loaded.
- Call `createImageBitmap` function in case of receiving `bitmapOptions` and supported by user browser, excluding Firefox and IE. 

### EXAMPLE OF USE:
- Forcing premultiply to true: 
`assetManager.loadTexture(texturePath, onSuccess, onFailed, {"premultiplyAlpha": "premultiply"})`
- Forcing premultiply to false: 
`assetManager.loadTexture(texturePath, onSuccess, onFailed, {"premultiplyAlpha": "none"})`

Also, these are other options supported by `createImageBitmap` that can be useful to users: 
- `imageOrientation`: Specifies whether the image should be presented as is or flipped vertically. Either none (default) or flipY.
- `premultiplyAlpha`: Specifies whether the bitmap's color channels should be premultiplied by the alpha channel. One of none, premultiply, or default (default).
- `colorSpaceConversion`: Specifies whether the image should be decoded using color space conversion. Either none or default (default). The value default indicates that implementation-specific behavior is used.
- `resizeWidth`: A long integer that indicates the output width.
- `resizeHeight`: A long integer that indicates the output height.
- `resizeQuality`: Specifies the algorithm to be used for resizing the input to match the output dimensions. One of pixelated, low (default), medium, or high.


badlogic:
Sorry for the late reply on this. The changes to AssetManager look good to me. I don't have a CLA on file from you yet. Could I bother you to follow the instructions in the "Contributing" section in README.md? I can then merge the PR into the 3.8 and 4.0-beta branches.

ivanpopelyshev:
Guys, do you know any production 2d renderer that relies on createImageBitmap already? 

We have regular problems with things like "colorSpaceConversion". If you use `createImageBitmap` then it wont be determined automatically like it does in simple `Image` case!

badlogic:
I did some more research. createImageBitmap() isn't supported on iOS either: https://caniuse.com/?search=createimagebitmap

This change is backward compatible, so it won't break existing code. However, I'm unsure if we should include something that doesn't for on the two other biggest browsers (Firefox/Safari).

Maybe this can be fixed another way? Do you have a test case for this @sp-jesus-diaz ?

badlogic:
I'm closing this out. Happy to re-open if you reply to my last question.

PR #1881, opened 2021-04-23
[BUGFIX]: mesh attachment squashing

aapavlov1994:
I have a bug with squashing mesh attachments when try to change animation state. it reproduced with

> animations made with spine v3.8.99
> pixi.js: 6.0.2
> pixi-spine: >= 2.1.10

So, i compared 2.1.10 and 2.1.11 versions of pixi-spine plugin and i think that found and fixed source of this bug.

**BUG representation:**

_expectation - 2.1.10 package and build with my fixes_
good

_reality - >= 2.1.10 packages_
bad


badlogic:
Uff, good catch! Thank you for the PR.

Issue #1273, opened 2019-02-08
[unity] Prefab instance always lists MeshFilter as override


HaraldCsaszar:
As reported on the Forum: 
http://esotericsoftware.com/forum/Unity-Nested-prefab-overrides-11476
http://esotericsoftware.com/forum/when-will-always-dirty-bug-be-fixed-16125

Problem seems to be in the double-buffered mesh that is assigned and changes again after being applied.

Resolution: Setting hide flags on the mesh did not yield the desired result in the prefab.
A workaround could be to always assign the first mesh uport save or clear the mesh every time before saving it.

ledvinam:
Possible solution could be to create mesh.asset which will be updated whenever we reexport spine projects but it won't set prefabs as changed.

HaraldCsaszar:
Creation of an additional `_Mesh.asset` file is not a viable solution here, we require every instance to show at least the first frame of the starting animation (not to speak of later showing an animation in Scene view). Having one mesh asset per instance is not acceptable for obvious reasons.

Setting `HideFlags` at MeshFilter might be promising. Unfortunately, at a quick test it caused indefinite reloading when selecting an already created prefab. Further investigations will be necessary.

kvfreedom:
Unity 2018.4.17: Selecting a prefab asset with a SkeletonRenderSeparator added in the project panel will load repeatedly.

HaraldCsaszar:
Thanks @kvfreedom for reporting, I have created a separate issue ticket for this issue here:
https://github.com/EsotericSoftware/spine-runtimes/issues/1626

ledvinam:
I will be happy to test this once it has been resolved. Please let me know.

ledvinam:
Harald, right clicking on the Sorting Layer label and click on Revert does not work.

HaraldCsaszar:
While setting `Fix PrefabMeshOverride` to `true` was working in Unity 2020.3, on Unity 2021.3 enabling `PrefabMeshOverride` seems to have no effect and still lists the prefab as having overrides.

HaraldCsaszar:
The previous fixes for this issue shall be re-evaluated and improved if possible, since the following forum threads show very high build time overhead due to cleanup of removing the procedural meshes:
http://esotericsoftware.com/forum/BuildPlayerWindow-OnGUI-mouseUp-during-build-WebGL-17603
[...]

PR #1938, opened 2021-08-06
[unity] Support fix prefab override via mesh filter global settings

vhristov:
Related to #1273


HaraldCsaszar:
Thanks for your contribution @vhristov. The pull request is now integrated in both 4.0 and 4.1-beta, with the improvement to also automatically provide backward-compatibility for already serialized scenes. New 4.0 and 4.1-beta unitypackages are available for download.

PR #1939, opened 2021-08-11
[unity] Fix multiple nested prefab override issue.

sr4dev:
Related to #1273.
http://ko.esotericsoftware.com/forum/Unity-Nested-prefab-overrides-11476


sr4dev:
`An asset is marked with HideFlags.DontSave but is included in the build:
Asset: 'Assets/xxxxxx/xxxxxxx.prefab'
Asset name: Skeleton Prefab Mesh "xxxxxxxxx"
(You are probably referencing internal Unity data in your build.)`

AssetBundle build faild with this message.
I'm finding a solution.

PR #1946, opened 2021-08-28
Fix 'Debug 64bit' build error on windows (VS2019)

aismann:
Error (active)	E0075	operand of '*' must be a pointer but has type "bool"


badlogic:
Cheers!

PR #1947, opened 2021-08-30
Fix command pipeline state uniform location mismatch

halx99:
This is a critical bug when there are more than 1 spine animations with different shader, please take a look


badlogic:
Cheers! Do you have a reproduction sample for me? I'd like to add it to the examples so this doesn't break in the future.

halx99:
Sure, I will provide for you later.

halx99:
Hi, here is the reproduction with adxe, just simple replace `cpp-template-default` of adxe, and test with cocos2d-x-4.0 is pointness, because only adxe have better custom shader support currently:
[cpp-template-default.zip](https://github.com/EsotericSoftware/spine-runtimes/files/7231026/cpp-template-default.zip)

## without patch:
without-patch-1
 without-patch-2

## with patch
The render result is correct as expected:
with-patch

halx99:
And for quick use spine runtime without this PR patched in `adxe`, you can simply copy files from https://github.com/adxeproject/spine-adxe/tree/adxe-1.0-b1/spine38 and to `adxe/extension/spine`

PR #1955, opened 2021-09-23
[ts] Made AnimationStateListener functions optional

jmlee2k:
All functions in AnimationStateListener are now optional - I believe this better reflects it's intended usage.

I've verified that the code checks for the existence of the functions before running them: https://github.com/EsotericSoftware/spine-runtimes/blob/4.0/spine-ts/spine-core/src/AnimationState.ts#L1042-L1075.

In addition, I've done some tests on some of our projects using pixi-spine, and everything works as expected.


badlogic:
Cheers!

PR #1965, opened 2021-10-11
[ts] SkeletonDebugRenderer fix in drawBones

tionkje:
skeleton.x and skeleton.y are already added in bone.worldX and
bone.worldY.
See Bone.ts => updateWorldTransformWith.


badlogic:
Ouch, good catch! Could I ask you to send us a signed a CLA via email? See https://github.com/EsotericSoftware/spine-runtimes#contributing

badlogic:
Cheers!

PR #1971, opened 2021-10-20
[cpp] Fix AABB calculation

Nitrillo:
The AABB calculation uses the wrong initial values, so the bounds are not calculated at all because new values are never bigger/lower than the initial values.


badlogic:
Oi, cheers!

dnd29:
Even after correction, it may be calculated incorrectly. I think it should be changed as follows. 

	float minX = FLT_MAX;
	float minY = FLT_MAX;
	float maxX = -FLT_MAX;
	float maxY = -FLT_MAX;

PR #1981, opened 2021-11-07
[unity] Fix skleton mesh "Identifier uniqueness violation" warning

### LINKED ISSUE #1973: [Unity] Prefab "Identifier uniqueness violation" warnings
Having multiple prefabs/prefab variants with SpineAnimation component on GameObjects with same name added to scene resulting in warnings, for example:
I have scene with prefab "Base" which contains GameObject "Human" with SpineAnimation, and prefab variants of base "Variant1" and "Variant2". This resulting in warnings:
`Identifier uniqueness violation: 'Name:Skeleton Prefab Mesh "Human", Type:Mesh'. Multiple Objects with the same name/type are generated by this Importer. There is no guarantee that subsequent imports of this asset will properly re-link to these targets.`
The issue is in SpineEditorUtilities.cs where name for mesh is generated from GameObject's name:
`string meshName = string.Format("Skeleton Prefab Mesh \"{0}\"", renderer.name);`
There should be some way to get rid of these warnings, maybe making mesh names more unique.

vhristov:
To remove the warning a new unique name is created for each mesh, where
an index is added after the renderer name.

Closes #1973

### DIFF
--- a/spine-unity/Assets/Spine/Editor/spine-unity/Editor/Utility/SpineEditorUtilities.cs
+++ b/spine-unity/Assets/Spine/Editor/spine-unity/Editor/Utility/SpineEditorUtilities.cs
@@ -110,6 +110,7 @@ namespace Spine.Unity.Editor {
 
 		public static bool SetupSpinePrefabMesh(GameObject g, UnityEditor.AssetImporters.AssetImportContext context)
 		{
+			Dictionary<string, int> nameUsageCount = new Dictionary<string, int>();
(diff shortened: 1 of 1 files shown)


vhristov:
@HaraldCsaszar can you check and say is this change fine, and if yes, shall I push it against 4.0 or 4.1 ?

HaraldCsaszar:
Thanks very much for providing the pull request! I will apply it to both 4.0 and 4.1-beta branches.

PR #1994, opened 2021-11-30
[Unity] Sprite Mesh Type Tight causes issues when used to replace attachments

### LINKED ISSUE #1884: [unity] Sprite Mesh Type `Tight` causes issues when used to replace attachments
One of the problems is when using `GetRemappedClone` with a `Sprite`, calling `ToAtlasRegion(true)` which does not respect the tight packing area.

Remark: calling `ToAtlasRegion(false)` which seems more appropriate does also not resolve the problem. UV coords shift in the opposite direction, making the problem worse instead of solving it. Shall be investigated.

This issue is also encountered on this forum thread:
http://esotericsoftware.com/forum/Replace-body-part-via-code-15764

whitecostume:
Fixes #1884

### DIFF
--- a/spine-unity/Assets/Spine/Runtime/spine-unity/Utility/AtlasUtilities.cs
+++ b/spine-unity/Assets/Spine/Runtime/spine-unity/Utility/AtlasUtilities.cs
@@ -198,13 +198,14 @@ namespace Spine.Unity.AttachmentTools {
 			Vector2 boundsMin = bounds.min, boundsMax = bounds.max;
 
 			// Texture space/pixel units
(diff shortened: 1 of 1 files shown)


HaraldCsaszar:
Thanks very much for your contribution! Closed with above commit 9ecf28b on 4.1-beta branch.

Note: We can't merge this to 4.0 since we need to be on the safe side for existing projects. Unity documentation about `Sprite.textureRect` states "Raises an exception if this sprite is tightly packed in an atlas.", although it's not raised in Unity 2017 nor latest Unity versions:
https://docs.unity3d.com/2022.1/Documentation/ScriptReference/Sprite-textureRect.html

PR #2000, opened 2021-12-09
[Unity] Fixed MaterialChecks to correctly check for Sprite material

MichalBerlinger:


MichalBerlinger:
Seems that now it unfortunately produces this when using shader Spine/Sprite/Unlit
Warning: 'Add Normals' required when not using 'Fixed Normals'!

Before the requested change it was producing:
Warning: Incorrect Texture / Material Settings found: It is strongly recommended to enable 'Alpha Is Transparency' on 'Straight alpha' textures. Assuming 'Straight alpha' texture because 'Straight Alpha Texture' is enabled at material).
Which was not correct, because I am using Spine/Sprite/Unlit shader with PMA Vertex, PMA Texture and the texture has alphaIsTransparency = false

So either MaterialChecks.IsSpriteShader() or MaterialChecks.RequiresMeshNormals() does its job wrong..

HaraldCsaszar:
Thanks for reporting. We've fixed this issue in commit 3025b37 above, both on 4.0 and 4.1-beta branches.
New 4.0 and 4.1-beta unitypackages have been released.

PR #2004, opened 2021-12-16
[ts] fix: set globalAlpha always

army8735:


badlogic:
What's the bug this is fixing?

army8735:
image
With this code:
```js
if (color.r != 1 || color.g != 1 || color.b != 1 || color.a != 1) {
  ctx.globalAlpha = color.a;
}
```
Ctx.globalAlpha has always been wrong 0.

image
It is correct after removing the if statement.

badlogic:
Cheers!

PR #2012, opened 2022-01-05
[spine-cpp] Improve RTTI compare performance when compiler string pooling enabled

halx99:
This patch will avoid string compare when compiler string pooling enabled, such msvc compiler flag: `/GF`

### DIFF
--- a/spine-cpp/spine-cpp/src/spine/RTTI.cpp
+++ b/spine-cpp/spine-cpp/src/spine/RTTI.cpp
@@ -36,6 +36,22 @@
 
 using namespace spine;
 
+static bool checkCompilerStringPooling() {
+	const char* l1 = "rtti";
+	const char* l2 = "rtti";
+	return l1 == l2;
+}
+
+static bool addressEqual(const char* lhs, const char* rhs) {
+	return lhs == rhs;
+};
+
+static bool stringEqual(const char* lhs, const char* rhs) {
+	return !strcmp(lhs, rhs);
+};
+
+static auto classNameEqual = checkCompilerStringPooling() ? &addressEqual : &stringEqual;
+
 RTTI::RTTI(const char *className) : _className(className), _pBaseRTTI(NULL) {
 }
 
@@ -47,13 +63,13 @@ const char *RTTI::getClassName() const {
 }
 
 bool RTTI::isExactly(const RTTI &rtti) const {
-	return !strcmp(this->_className, rtti._className);
+	return classNameEqual(_className, rtti._className);
 }
 
 bool RTTI::instanceOf(const RTTI &rtti) const {
 	const RTTI *pCompare = this;
 	while (pCompare) {
-		if (!strcmp(pCompare->_className, rtti._className)) return true;
+		if (pCompare->isExactly(rtti)) return true;
 		pCompare = pCompare->_pBaseRTTI;
 	}
 	return false;


badlogic:
I doubt this has any meaningful impact. First, strcmp already does a pointer check on most platforms as an early out. Second, you are assigning a function pointer to a static based on your compiler feature detection. Instead of being able to inline the strcmp call, we now have to resolve a function pointer and call the function without inlining.

halx99:
For literal strings, the strcmp pointer check not enough, because when two literal string address not equal,  the strcmp will always do characters compare   loop and the performance is slow than  just check address equal or not.

badlogic:
Of course, but the RTTI string literals are instantiated once for each type, stored statically, and are never compared with string literals not generated via the RTTI macros. So the compiler can actually inline a simple pointer comparison. Whereas with your patch, the compiler must always fetch the function pointer and perform a proper function call.

halx99:
I doubt, I will check the release build asm code to verify the inline behavior described by you.

badlogic:
Here: https://godbolt.org/z/7KW6Ken34

Clang is actually smart enough to not call `RTTI::instanceOf` at all and simply removes the conditional in `main()` with `-O3`
MSVC doesn't call into `strcmp` at all with `/O2`

As I said, our current RTTI implementation is pretty optimal when it comes to `instanceOf()` on all the compilers that are relevant for us.

halx99:
> Here: https://godbolt.org/z/7KW6Ken34
> 
> Clang is actually smart enough to not call `RTTI::instanceOf` at all and simply removes the conditional in `main()` with `-O3` MSVC doesn't call into `strcmp` at all with `/O2`
> 
> As I said, our current RTTI implementation is pretty optimal when it comes to `instanceOf()` on all the compilers that are relevant for us.

Got it, thanks and happy to know these compilers so much smart.

halx99:
Updated:

I have test on the game engine project: it's seems the msvc with /O2 just inline strcmp code, see below screenshot:

image

halx99:
Use constexpr for line, the compiler can generate more simple `asm` code
```cpp
template<bool = true>
struct LiteralCompare
{
	inline static bool equals(const char* lhs, const char* rhs) { 
		return lhs == rhs;
	}
};

template<>
struct LiteralCompare<false>
{
    inline static bool equals(const char* lhs, const char* rhs) { 
		return !strcmp(lhs, rhs);
	}
};

constexpr bool checkCompilerStringPooling() {
	constexpr const char* l1 = "rtti";
    constexpr const char* l2 = "rtti";
	return l1 == l2;
}

RTTI::RTTI(const char *className) : _className(className), _pBaseRTTI(NULL) {
}

RTTI::RTTI(const char *className, const RTTI &baseRTTI) : _className(className), _pBaseRTTI(&baseRTTI) {
}

const char *RTTI::getClassName() const {
	return _className;
}

bool RTTI::isExactly(const RTTI &rtti) const {
    return LiteralCompare<checkCompilerStringPooling()>::equals(_className, rtti._className);
}

bool RTTI::instanceOf(const RTTI &rtti) const {
	const RTTI *pCompare = this;
	while (pCompare) {
		if (pCompare->isExactly(rtti)) return true;
		pCompare = pCompare->_pBaseRTTI;
	}
	return false;
}
```
And the `isExactly` asm code is:

image

halx99:
So, don't rely on compiler optimizing, use enum to instead maybe better, just like `spine-c`

badlogic:
Even better: cam short circuit the conditional calling strcmp by first checking if the two class names are pointing to the same address.

PR #2018, opened 2022-01-13
[unity] BoneFollower components change Transform scale regardless of settings

MedMM:
Fixed bug with impossibility to change scale with BoneFollower component


MedMM:
Please, remove this hardcode and why do you always update local scale, even when you don't need it

HaraldCsaszar:
Your provided implementation is incorrect in two out of three code branches below your modifications (the `followLocalScale` and `followSkeletonFlip` branches). We will modify the code properly so that scale remains unchanged when all scale-related follow properties are disabled.

HaraldCsaszar:
Thanks for reporting. This has been fixed for `BoneFollower` and `BoneFollowerGraphic` components on the 4.1-beta branch. Not applying fix on 4.0 branch to not break behaviour at existing projects.

PR #2019, opened 2022-01-13
feat (loadTextureAtlas) add option to use file alias

MikalDev:
For Construct 3, in preview mode, the texture file names are changed to blobs with completely different filenames than what is listed in the atlas. So, loadTextureAtalas will not work because the blob file names don't match the page.name.  When Construct 3 is not in preview mode, the normal file names will match the page.name.

To get around this issue, I added an optional parameter to loadTextureAtlas, the object fileAlias, this will allow the Construct 3 plug-in to create a fileAlias object which maps between page.name and the C3 project file names, whether they are aliased or not. If this is not needed, the function performs as it did before.

### DIFF
--- a/spine-ts/package-lock.json
+++ b/spine-ts/package-lock.json
@@ -7949,41 +7949,41 @@
     },
     "spine-canvas": {
       "name": "@esotericsoftware/spine-canvas",
-      "version": "4.0.16",
+      "version": "4.0.17",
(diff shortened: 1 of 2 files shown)


badlogic:
Looks good, cheers! I'll do a release to NPM asap. Have to fix a few more things before that.

PR #2041, opened 2022-03-11
remove strdup(), use common MALLOC_STR instead, for better tracking m…

DimonSE:
Hi, I try use my custom memory allocator for spine-c/spine-cpp runtimes and found this old strdup() function. 

The problem with strdup is create untracked memory block inside strdup(), but we still need use FREE(self->hash); later and have problem with custom MALLOC/FREE, cause mismatch strdup()/custom FREE

I rewrite this probably old core, to use common MALLOC_STR


badlogic:
Good find! I've fixed this in both the 4.0 and 4.1-beta branch.

PR #2051, opened 2022-03-30
[ts] Fixed bug of reading Unicode from binary data

Jerry-Terrasse:
In file `spine-ts/spine-core/src/SkeletonBinary.ts`, line `1025`:

```typescript
let b = this.readByte();
switch (b >> 4) {
    ...
}
```

When dealing with Korean or Chinese, `b` can be some thing like `0xe4`, which is regarded as a negative number.
Thus, `b>>4` equals to `-2` instead of `14`, causing error `Region not found in atlas: *** (mesh attachment:  ***)`.


badlogic:
Good find, thank you!

PR #2064, opened 2022-04-06
Update Skin.cs

DabManTrips:


HaraldCsaszar:
Unfortunately we can't provide a `RemoveSkin` method in a symmetrical way [1], and we don't want to provide an unintuitive `RemoveSkin` with unexpected behaviour.

Feel free to modify your local copy of the spine-unity runtime according to your needs though.
BTW: Your implementation is incorrect, it clears all bones an constraints!

[1] `addSkin(x)` and `removeSkin(x)` is not a no-op, e.g. in `addSkin(abc); addSkin(bcd); removeSkin(abc);`

PR #2083, opened 2022-05-18
[spine-c] Add initialization of mixRotate

DreadPirateSergey:
You probably forgot to add initialization of mixRotate.


badlogic:
Uff, good catch, cheers!

PR #2085, opened 2022-05-19
[ts] Use existing, explicit atlas data instead of relying on an image's current width/height

Jimbly:
This fixes a bug where UVs would sometimes be wrong if an image's size is not the same as when originally exported.  This could happen in the wild to _any_ project, due to mobile ISPs/browsers that proxy requests and resize images before sending them to clients.
Fixing this also allows `spine-core` to work with engines that:
* use embedded, low-res textures while waiting for high res textures to load
* automatically provide lower resolution textures to low-end devices or low-resolution displays (assuming care is taken with atlas padding upon export)

The issue can be trivially reproduced by resizing `spine-ts/spine-webgl/example/assets/spineboy-pma.png` (down or up, for example to 512x128), and running the WebGL example.  With this fix, the spine runtime generates the same texture coordinates regardless of what size of image the browser happens to load.

Question: I saw references to "non-essential data", is there any chance this data is not available in some circumstances?  I tried all examples and it seems to work fine (as well as fix the actual bug I was encountering while integrating it into my engine).

Possible improvement (though my TypeScript-fu is weak): Probably don't need to declare `getImage()` in `abstract class Texture` in `Texture.d.ts`, nor `_image` nor passing an HTMLImageElement to the constructor, however all of the official consumers of spine-core seem to use these, so maybe best to leave it alone.

### DIFF
--- a/spine-ts/spine-core/src/attachments/MeshAttachment.ts
+++ b/spine-ts/spine-core/src/attachments/MeshAttachment.ts
@@ -93,8 +93,8 @@ export class MeshAttachment extends VertexAttachment implements HasTextureRegion
 		let n = this.uvs.length;
 		let u = this.region.u, v = this.region.v, width = 0, height = 0;
 		if (this.region instanceof TextureAtlasRegion) {
-			let region = this.region, image = region.page!.texture!.getImage();
-			let textureWidth = image.width, textureHeight = image.height;
+			let region = this.region, page = region.page;
+			let textureWidth = page.width, textureHeight = page.height;
 			switch (region.degrees) {
 				case 90:
 					u -= (region.originalHeight - region.offsetY - region.height) / textureWidth;


badlogic:
Non-essential data is an export setting when exporting .json or .skel files and controlled by Spine Editor users. For production deployments, non-essential data should not be exported, as it can bloat the file size. It should not affect your proposed code change though, as the texture atlas page width and height are not part of the non-essential data.

I only had a brief look at your proposed changes and they look OK to me. However, I have to check if other parts of the rely on the page image's width/height instead of what's stored on the page directly.

Jimbly:
Great, thanks for taking a look.  That's the only function in `spine-core` referencing `getImage()`, so is the only place that the texture's image's width/height could be accessed, so I think it should have no effects outside of that function.  Architecturally it certainly makes sense that the spine runtime should be able to generate UVs without needing the pixel data of a texture.  Unless I'm missing something, the w/h specified in the page's data in the atlas should always be equal to the w/h of the image exported, unless something messes with the image in the meantime =).

badlogic:
In some use cases, people may intercept thr atlas page image loading and load a 2x or 0.5x image instead, depending on the device. Since texture coordinates are given in normalized form, we need to know the actual image dimensions. Hrm, actually writing that out, that shouldn't matter. I'll try to get to this next week!

Jimbly:
I think this is exactly the fix required to allow people to do that =)!  By using the texture width/height that was used to generate the region/etc coordinates, instead of the size of the image loaded, you can load whatever size image you want, and still get the same normalized coordinates back.

badlogic:
Sorry, totally forgot about this! Looks good.

PR #2087, opened 2022-05-30
EventTimeline::getFrames returns const & to avoid copying.

dumganhar:


badlogic:
Oh my, that's a good fix. Thanks!

PR #2096, opened 2022-06-21
[ue4] Black Spots and Normal Flipping Bug Fixes By CCW Resolve (Old Title : Normal Generation Improvement)

GGgRain:
All the vertex normals now have its own face direction. Old method creates only one vector and applied it on the whole vertices of the mesh. (Basically, all the triangles had the same normal direction before)
It will prevent some of the flickers in some occasions.
This will improve the experiences on modifying some bones' location with the bone drivers when they work with lit material.

Edit : Some of the flickers still appear. I'm now figuring out how to fix those too. Maybe giving some offsets to each vertices will help it.

Edit + : I found It directly causes some flickers that are never happened before. I'm checking out what is causing it, ans how to fix it.
So I will just check it to the draft until I find a solution.

**Before**

before

**After**

after

### DIFF
--- a/spine-ue4/Plugins/SpinePlugin/Source/SpinePlugin/Private/SpineSkeletonRendererComponent.cpp
+++ b/spine-ue4/Plugins/SpinePlugin/Source/SpinePlugin/Private/SpineSkeletonRendererComponent.cpp
@@ -350,7 +350,7 @@ void USpineSkeletonRendererComponent::UpdateMesh(Skeleton* Skeleton) {
 			indices.Add(idx + attachmentIndices[j]);
 		}
 
-		//Smooth normal calculation.
+
 
 		//Calculate total triangle to add on this loof.
 
@@ -358,17 +358,9 @@ void USpineSkeletonRendererComponent::UpdateMesh(Skeleton* Skeleton) {
 
 		int TriangleToAddNum = indices.Num() / 3 - TriangleInitialCount;
 
-		//Add empty normals for that represent all the vertices we are going to add for this mesh.
+		int FirstVertexIndex = vertices.Num() - numVertices;
 
-		for (int j = 0; j < numVertices; j++) {
-
-			normals.Add(FVector::ZeroVector);
-
-		}
-
-		FVector normal = FVector(0, 1, 0);
-
-		//loof through all the triangles and add a result of the cross product to the consisting vertice's normal of the triangles.
+		//loof through all the triangles and resolve to be reversed if the triangle has winding order as CCW.
 
 		for (int j = 0; j < TriangleToAddNum; j++) {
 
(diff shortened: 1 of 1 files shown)


badlogic:
That looks mostly OK. I wonder if we actually need such a complex solution. All faces in the skeleton share the same normal, which is basically the "look at" vector of the renderer component in the scene graph. Maybe we can just figure out that look at vector and use it as the normal for all faces instead?

GGgRain:
It is quite a necessary function.

It doesn't have anything to do when we use regionattachment, but when we use meshattachments with vertices that weights applied on it, there is a ton of possibilities that any triangle in the mesh can be flipped over and have the opposite direction from the others.

This is one of the cases :
http://ko.esotericsoftware.com/forum/ue4-Lit-material-is-broken-somehow-17027

And it can be simply fixed by this code, since it will calculate the perfect normals for each vertices.

GGgRain:
Okay, I finally came up with a more clever idea, and I also found out there were some unanswered questions from your last comment.

_**which is basically the "look at" vector of the renderer component in the scene graph. Maybe we can just figure out that look at vector and use it as the normal for all faces instead?**_

Well, It turns out that your approach was right in some ways, even though the solution was quite off from what we expected to see. 

The problem is, if we generate identical normals for every vertices, then some of the triangles that might be facing backward of the normal vector direction will not be rendered correctly. This happens because each triangles have own surface normal that is decided by the vertices winding order, and we must provide the vertices normal set that contains data about the near triangle's facing direction, but if we give just one single vector, it will not fit to some of the triangles.

So my initial idea for my approach for this bug was changing each normals to fit with the triangles, but it turned out it is very complex and still left with some un-avoidable black spots on its result.

And Finally, I changed the approach this time.
Instead of generating normals to fit with the direction of the triangles,
I changed the triangle itself to make it fit with the normals, and It works like charm.

(Sorry if the sample image seems quite disturbing... I intentionally strecthed some joints to show you it is being rendered well without the ugly black spots)
캡처

I will commit my code by today, and please check again for this request!

badlogic:
That's a good solution! Could I bother you to sign and send me our contributor license agreement to contact@esotericsoftware.com? Otherwise we can't publish your code ourselves. http://esotericsoftware.com/licenses/cla.txt

GGgRain:
I sent it. The name of the mail is "Agreement for the contribution for providing a pull request : https://github.com/EsotericSoftware/spine-runtimes/pull/2096"

I hope to see that the ue4 version of the runtime plugin be more practical to everyone!
Of course any users like me can make a small contribution just like what I did today, but I'm really looking forward to see the official update from your company and hopefully it improves the experiences with the plugin someday! (The last update was.... almost 1~2 years ago if we don't count bug fixes or minor changes.. and due to that, I needed to implement all the things just like this...)

GGgRain:
Umm, I wonder why this pull request is still not patched on to the code. Is there any problem? Or should I do something to patch it? (It's basically my first pull request on the github so i'm confused)

badlogic:
Sorry, I haven't had time yet, as I have to post process it manually (formatting) and make time to test the changes. I'll try to get to it this week.

GGgRain:
Ok, great! I feel relieved because you said you don't have any problem wifh the code! Have a time and thank you for your work on this.

badlogic:
I have now merged and tested your PR. Works great. Thanks!

PR #2099, opened 2022-06-25
Fix for Linux build: conversion from 'long long int' to 'const Variant' is ambiguous

bryab:
Was getting build error on Linux/GCC:

```
spine-runtimes/spine-godot/spine_godot/SpineTimeline.cpp:91:40: error: conversion from 'long long int' to 'const Variant' is ambiguous
   91 |   result[i] = (spine::PropertyId) ids[i];
```

I'm not really a C++ dev so not sure if what I've done here is correct, but this resolves my build issue.

GCC version:
`gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0`


badlogic:
Cheers!

PR #2104, opened 2022-07-05
[cocos2d-x] Use backend shader definations

halx99:


badlogic:
Cheers! Did the uniform and attribute names change? The code still works with the latest 3.x and 4.x cocos2d-x releases.

halx99:
Yes, the code compatible with cocos2d-x, and the uniform and attribute names were changed in a fork of cocos2d-x `axmolengine`.

PR #2114, opened 2022-07-15
[UE4] Prevent a crash when a skin has no attachments.

torus:
When importing a JSON file with an empty skin, the importer crashes.
This patch adds a NULL check so that the importer just skips the attachments when the skin doesn't have attachments.

We are still using 4.0, so this PR is for the 4.0 branch.


badlogic:
Cheers!

Issue #2065, opened 2022-04-07
[cpp] Path and/or PathConstraint not working


badlogic:
See http://esotericsoftware.com/forum/GameMaker-Studio-2-issues-with-Spine-file-17245?p=76158#p76158

PR #2115, opened 2022-07-15
[c] Fixed JSON parsing of PathConstraint spacing mode (#2065).

yoyo-danielc:
See 4ec3290b987230a0b71d67c34269ba50f4e5e57f

(#2065 was fixed in spine-cpp, but not spine-c)


badlogic:
Thanks!

PR #2122, opened 2022-07-20
Corrected Tint black

PolRecasensSarra:
I was using spine with Godot and the tint black property in the Godot Editor wasn't working correctly because the red channel was discarded and the green channel was being used instead.

This is the result color before the fix.
before

And this is the result color after the fix.
after


badlogic:
Great (and embarassing) find, thanks! That said, the Godot runtime currently doesn't support tint black, as that'd require a custom shader. Did you write your own tint black shader?

PolRecasensSarra:
Yes, I have my own tint black shader, but unfortunately I can't share it because it belongs to my company and it's under NDA.

badlogic:
Super cool! And totally understand you can't share the shader. But could you tell me how you are pushing the dark color to the vertex shader through the `VisualServer/RenderingServer` API? Do you pack it into the bones or weights array passed to `canvas_item_add_triangle_array `? Does that work on mobile? I figured those would be int16/fp16 on mobile for example.

PR #2139, opened 2022-08-19
[ts] Fix wrong null check on ShapeRenderer polygon and curve functions

davidetan:
I was playing with the WebGL demos and I found that when Strict Mode was enabled during [this commit](https://github.com/EsotericSoftware/spine-runtimes/commit/e4372e5329f9fe66b7b97e939d2fa800b04ce99a#diff-9b2ef4add9c3f30d741a3184c535e675cfdfc37b9d0e1975416099cc1070ad37L261-R261) a pair of null check was not translated correctly in 

This lead vine and tank examples to crash when the checkbox to show the paths is checked. The fix should solve the problem.


badlogic:
Ouch, good catch! I've fixed this in 4.1 and 4.2-beta and published new packages to NPM.

PR #2152, opened 2022-09-13
fix error C4146 in MSVS UWP

DimonSE:
Hello

I have strange error in Visual Studio 2019 v16.11.18
sometimes it constantly reproduced, sometimes go away

but anyway it's code look like really risky
Can you pls check this moment, maybe we have a bug here

For temporary fix I just restore (int) from v4.0

here is compile error, problem in -(value & 1)
lib\spine\SkeletonBinary.c(131,64): error C4146: unary minus operator applied to unsigned type, result still unsigned

Best regards,
Dmitriy Sechin | Programmer

### DIFF
--- a/spine-c/spine-c/src/spine/SkeletonBinary.c
+++ b/spine-c/spine-c/src/spine/SkeletonBinary.c
@@ -128,7 +128,7 @@ static int readVarint(_dataInput *input, int /*bool*/ optimizePositive) {
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


badlogic:
4.0 doesn't cast to `int` either:
https://github.com/EsotericSoftware/spine-runtimes/blob/4.0/spine-c/spine-c/src/spine/SkeletonBinary.c#L130

This seems to be an errornous port. It will work on pretty much all C compilers, but the correct fix is to use binary negation via `~`. I've fixed this in the 4.1 and 4.2-beta branches. Thanks for reporting!

PR #2160, opened 2022-09-27
 [unity] Fixed excessive UnloadingUnusedAssetsImmediate during build.

shniqq:
Instead of calling `EditorUtility.UnloadUnusedAssetsImmediate()` for each prefab in the project, it is now only called once. 

In our project this reduced build times by ~3 minutes, and it was also the main culprit for long build times (iOS export). We have around 2300+ prefabs.

In my testing this resulted in the same build as calling it for each prefab.

Before:
image
After:
image


HaraldCsaszar:
Thanks for reporting @shniqq, we're sorry for the long build time overhead.

HaraldCsaszar:
Please also note that you can disable preprocessing the prefabs entirely via `Edit - Preferences` - `Spine`.

HaraldCsaszar:
A new 4.1 spine-unity package containing the fix is available for download here as usual:
http://esotericsoftware.com/spine-unity-download

PR #2183, opened 2022-10-21
Fix webglConfig check

WolfgangKurz:
Missing negative operator i guess.


badlogic:
Ouch, good catch. Cheers!

Issue #2198, opened 2022-11-22
[spine-cpp] cpp version displayed not as expected with editor or ts version


ggggamer:
This is `cpp` version：
![cpp version](https://user-images.githubusercontent.com/4241137/203236927-4113ad5c-6f69-4cfa-aa6e-3e1e303742f7.png)

Below is it displayed in `ts` version，same as in editor or `skeletonViewer-3.8.99.jar`
![editor or ts version](https://user-images.githubusercontent.com/4241137/203236975-12a334ec-3156-4f9c-a205-1a58b1575cae.png)

Both using current `3.8` branch.

[project.zip](https://github.com/EsotericSoftware/spine-runtimes/files/10064357/project.zip)
[exported.zip](https://github.com/EsotericSoftware/spine-runtimes/files/10064358/exported.zip)

PR #2199, opened 2022-11-23
fix IKConstraint apply error in spine-cpp

ggggamer:
#2198  Compare ts and cpp version, this line is different, the issue is fixed after we swap this line.

image

This is fixed by [@kenshin](https://forum.cocos.org/u/huanxinyin/summary) at forum.cocos.org


badlogic:
Good catch. I've applied a fix manually. Note that the spine-libgdx runtime is the reference runtime, not spine-ts.

PR #2205, opened 2022-11-25
fix incorrect ordering issue when applying custom materials to slot n…

### LINKED ISSUE #2195: [godot] SpineSlotNode applies normal material to wrong node if Draw Order is changed during animation.
SpineSlotNode applies normal material to wrong node if Draw Order is changed during an animation.

https://user-images.githubusercontent.com/78934401/202018454-6a3a5e22-7433-41de-8ecc-0726e3628583.mp4

image

jordo:
fixes https://github.com/EsotericSoftware/spine-runtimes/issues/2195


badlogic:
Thanks!

PR #2211, opened 2022-12-07
[flutter] Fix linting and code formatting errors

jtakakura:


badlogic:
I know pub.dev is very peculiar when it comes to formatting and demands a line width of 80 characters. They will also deduct points from the score if we don't adhere to that. However, we'll stick with our setting of 120 characters per line.

As for the linting errors: weird, shouldn't dart analyze show that? I run that before commits now and the output is clean. How do you run the linter?

I'd be happy to merge the linter commit.

jtakakura:
This lint error occurred after formatting the code.
Then it's no need to merge my pr if you keep your formatting rule.
Sorry for being unclear.
I close this pr.

PR #2220, opened 2022-12-30
[spine-ts][player] Fix skeleton bounds now incorporate clippings

### LINKED ISSUE #2219: [player] Take clipping into account when calculating animation viewport
https://github.com/EsotericSoftware/spine-runtimes/blob/6661ae9322986c84267f7b4b71d66d332dad6d29/spine-ts/spine-player/src/Player.ts#L762

I suggest to calculate viewport size by clipping, but not min-max vertices. Sometimes animation goes out of bounds and player fits viewport size to not visible parts. It's unexpected behaviour for me, so I have to setup viewport manually which is taking much time and exhausting. 
See example, I expect to see these two animations in same bounding box, but 2nd orange jump outside of bottom border

(Could try to provide PR later if you agree with idea. But only for JS platform)

image
image

Disorrder:
Fixes #2219 
Calculate only bounding of clipping attachments, if they exist. 
Example: Violet border is debug of clipping

**Before:**
image
**After:**
image

Spineboy
**Before**
image
**After**
image

### DIFF
--- a/spine-ts/spine-core/src/Skeleton.ts
+++ b/spine-ts/spine-core/src/Skeleton.ts
@@ -27,6 +27,7 @@
  * THE SPINE RUNTIMES, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *****************************************************************************/
 
+import { ClippingAttachment } from "./attachments";
 import { Attachment } from "./attachments/Attachment";
 import { MeshAttachment } from "./attachments/MeshAttachment";
 import { PathAttachment } from "./attachments/PathAttachment";
@@ -610,6 +611,7 @@ export class Skeleton {
 	getBounds (offset: Vector2, size: Vector2, temp: Array<number> = new Array<number>(2)) {
 		if (!offset) throw new Error("offset cannot be null.");
 		if (!size) throw new Error("size cannot be null.");
+		let onlyClippings = false;
 		let drawOrder = this.drawOrder;
 		let minX = Number.POSITIVE_INFINITY, minY = Number.POSITIVE_INFINITY, maxX = Number.NEGATIVE_INFINITY, maxY = Number.NEGATIVE_INFINITY;
 		for (let i = 0, n = drawOrder.length; i < n; i++) {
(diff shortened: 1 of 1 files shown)


Disorrder:
Not sure it is totally correct, because in spineboy animation clipping applies only for boy, but portal is visible and should be included in bounding. But idk how to recognise clipped components
image

badlogic:
Ah, yes, that won't work I'm afraid. What needs to be done is replicating large parts of the `SkeletonRenderer` logic:
https://github.com/EsotericSoftware/spine-runtimes/blob/4.1/spine-ts/spine-webgl/src/SkeletonRenderer.ts#L65

The bounds need to be evaluated by getting the min/max x/y for clipped region and mesh attachments. Basically, the `SkeletonRenderer.draw()` method needs to be duplicated. That shouldn't be done in `getBounds()` I think, but rather in the Player itself. Other people's projects rely on `Skeleton.getBounds()` not taking into account clipping attachments.

Disorrder:
Thanks for fast replies!
But viewport size also relies on `Skeleton.getBounds()`. And any other projects could rely on this behaviour 🤔 . Do you think it should use new bounding method? Like getBoundsVisible or something?
Did I understand you correct? Viewport size should be calculated from clipped regions and only _visible_ meshes that aren't clipped.
I don't think we really need to repeat draw method, just need to filter all attachments that aren't clipped and clipping regions itself, that's it. But I still not sure how to recognise is element clipped or not. Are `slotRangeStart `, `slotRangeEnd` responsible for that? Multiple clip regions are not supported, right? Or they just somehow sorted 🤔 
I've just started to use your player couple days ago, haven't dived too deep yet.

davidetan:
@Disorrder Thanks for you PR and the discussion.
This inspired us to change `Skeleton.getBounds()` in all our runtimes.
Now the method gets an additionional `SkeletonClipping` parameter. If passed, the bounds are calculated taking clipping attachments into consideration.
You can see the updates to the Player and TS runtime here: 03f8f67 and f309722.

PR #2236, opened 2023-02-03
[godot] Update parse_property signature for Godot 4

karljs:
This gets the Godot 4 plugin building correctly again after a signature change.


badlogic:
Huh, I already had this fixed. Wonder where the commit went :(

Thanks for the PR!

badlogic:
Ah, I fixed it in the 4.2-beta branch...

PR #2270, opened 2023-04-04
[spine-c] Fix crash when skeleton `hash` or `spine` doesn't exist in JSON file

AGulev:
If String value doesn't exist, `Json_getString(Json *json, const char *name, const char *defaultValue)`  returns `defaultValue` which is `const char*`. 
In such case, using `0` like here

```
Json_getString(skeleton, "hash", 0);
```
`Json_getString()` will return `null` pointer, which will crash in `MALLOC_STR(TO, FROM)` macros on `strlen(FROM)`.

To avoid it, here should be used a default string value, like `"0"`.

(I've sent CLA)


badlogic:
Got catch, sorry it took so long for me to get to this!

AGulev:
No worries, I fully understand.
Thank you!

PR #2295, opened 2023-05-27
[libgdx] fix setting the color format of packedColor and packedDarkColor in TwoColorPolygonBatch. 

qinxiandiqi:
The color format of packedColor and packedDarkColor is abgr8888, not rgba888.


badlogic:
Cheers!

PR #2299, opened 2023-06-03
feat(spine-ts): add support for pixi.js

miltoncandelero:
Hello!

This pull request adds support for Pixi.js, to the spine-ts runtimes. 

The main idea behind this is to be feature complete and as out-of-the-way as possible. This pr does not aim to re-implement anything already existing in either spine-ts or in pixi.js

To achieve this, the renderer un-interleaves the geometry that spine-ts generates so that pixi renderer can interleave it again. 

Bone transformation is calculated by spine-ts, skeleton position on screen and posterior transformations are handled by pixi.

This code is currently being battle tested at my daytime job at Killabunnies to develop 3 games (I can't give more details about those games yet, sorry 😅 )

## Key features:
1. Rendering
   - Rendering is handling by creating one pixi mesh per Region or Mesh attachment in a flat structure. Pixi renderer will interleave and batch these meshes along with all other sprites and meshes in a scene due to it's multi-texture batching system.
   - Dark tint is supported and batched, however it can't be batched with regular sprites or meshes and needs a render pass of it's own. That is why it's disabled by default and if detected that is needed a warning will be thrown to explain the user how to enable it correctly.
   - The Mesh class used internally, can be changed (this is how dark tint is achieved). This allows users to write shaders that integrate illumination or normal maps.
   - Polygon packed atlas are fully supported
   - Pixi meshes used for rendering are pooled and try to instantiate as little memory as possible

2. Assets
   - Assets are to be loaded via Pixi's `Asset` class and you must load both the skeleton and the atlas file. We will not try to guess the atlas corresponding to a skeleton. This allows for greater flexibility when using a single skeleton with different textures.

3. Ergonomics
   - To simplify creation there is a helper static method `Spine.from(...)` that creates the skeleton data and tries to cache it, similar to what the Phaser runtime does. The `Spine` constructor works similar to the constructors of other spine-ts implementation, requiring an `SkeletonData`
   - AutoUpdate is on by default, but can be turned on during construction or even after construction. Manually updating an autoupdate animation will trigger a warning.
   - Functions to get a bone position and to set a bone position were added.

4. Debug
   - A debug renderer tool is included. It can use pixi objects to render different debug layers, similar to the SpineViewer.jar

5. Events:
   - Pixi.js currently has an issue with custom events: https://github.com/pixijs/pixijs/issues/8957 so the events are not forwarded from spine's original events. For now it is advised to connect directly to the events in the `state` variable.

## Things missing
- More examples
- Better inline/generated documentation
- I am not sure if the toolchain/code style is compliant with the spine-ts way.


badlogic:
This is great! The polypack atlas support seems to have an issue though. This is what I get on latest Chrome on macOS, but I guess it's an issue related to generating regions out of the polypack atlas.

Screenshot 2023-06-07 at 14 37 34

Should look like this, and is what you get when using the Spine atlas instead.

Screenshot 2023-06-07 at 14 41 47

This is 100% in line with what I planned myself, so that's great! I'll do a formatting pass and code review and will add docs and examples as appropriate.

Thanks a ton! Before I can merge this, I'll need a CLA, so we can ship it under the Spine Runtimes license. See [Contributing](https://github.com/EsotericSoftware/spine-runtimes#contributing) at the bottom of the README.md. Sorry for the red tape :(

badlogic:
Since I'm not a Pixi export, I'm leaving a few questions/comments on the files in this PR. Hope you don't mind.

spine-ts/spine-pixi/src/SlotMesh.ts
```diff
@@ -0,0 +1,89 @@
...
+			console.warn("DarkTint is not enabled by default. To enable use a DarkSlotMesh factory while creating the Spine object.");
+			this.warnedTwoTint = true;
+		}
+
+		SlotMesh.auxColor[0] = finalVertices[2];
+		SlotMesh.auxColor[1] = finalVertices[3];
+		SlotMesh.auxColor[2] = finalVertices[4];
+		SlotMesh.auxColor[3] = finalVertices[5];
+
+		this.tint = SlotMesh.auxColor;
```
badlogic, line 72: Since the mesh tint isn't part of the vertex data, I assume it's passed as a uniform? If so, how can Pixi batch meshes with different tints? I couldn't find anything on that on the web.

badlogic:
Opened a TODO for me here: https://github.com/EsotericSoftware/spine-runtimes/issues/2305

spine-ts/spine-pixi/src/SpineTexture.ts
```diff
@@ -0,0 +1,109 @@
...
+
+	public setWraps(uWrap: TextureWrap, _vWrap: TextureWrap): void {
+		this.texture.baseTexture.wrapMode = SpineTexture.toPixiTextureWrap(uWrap);
+
+		// Pixi only has one setting
+	}
+
+	public dispose(): void {
+		// I am not entirely sure about this...
+		this.texture.destroy();
```
badlogic, line 38: Textures are cached in the static `textureMap` field. I suppose on `dispose()` the texture should get removed from the cache? I'm not familiar with the Pixi resource life cycle.

spine-ts/spine-pixi/src/SpineTexture.ts
```diff
@@ -0,0 +1,109 @@
...
+		// Todo: maybe add error handling if you feed a video texture to spine?
+		super((image.resource as BaseImageResource).source as any);
+		this.texture = PixiTexture.from(image);
+	}
+
+	public setFilters(minFilter: TextureFilter, _magFilter: TextureFilter): void {
+		this.texture.baseTexture.scaleMode = SpineTexture.toPixiTextureFilter(minFilter);
+		this.texture.baseTexture.mipmap = SpineTexture.toPixiMipMap(minFilter);
+
+		// pixi only has one filter for both min and mag, too bad
```
badlogic, line 27: Ouch. Using the min filter is the sensible choice then. Should be documented in the docs on the website later.

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+import type { SpineTexture } from "./SpineTexture";
+import { SlotMesh } from "./SlotMesh";
+import type { ISpineDebugRenderer } from "./SpineDebugRenderer";
+import { Assets } from "@pixi/assets";
+import type { IPointData } from "@pixi/core";
+import { Ticker, utils } from "@pixi/core";
+import type { IDestroyOptions, DisplayObject } from "@pixi/display";
+import { Container } from "@pixi/display";
+
+export interface ISpineOptions {
```
badlogic, line 26: We generally do not prefix interfaces with `I`. Not something you need to fix, I'll go through it with our style comb once I merged it.

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+	removeUnusedSlots?: boolean;
```
badlogic, line 27: When would this be useful? A skeleton generally has a fixed number of slots. The maximum number of `SlotMesh` instances would thus be the number of slots in the skeleton. Does recycling have any benefit?

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+				end: (trackEntry) => this.emit("end", trackEntry),
+				event: (trackEntry, event) => this.emit("event", trackEntry, event),
+				interrupt: (trackEntry) => this.emit("interrupt", trackEntry),
+				start: (trackEntry) => this.emit("start", trackEntry),
+			};
+			this.state.addListener(this.spineListeners);
+		*/
+	}
+
+	public update(deltaSeconds: number): void {
```
badlogic, line 119: This is only called in the auto-update case, correct? If so, then is the user supposed to call `update()` manually in the non-auto-update case?

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+			Spine.clipper.clipEndWithSlot(slot);
+		}
+		Spine.clipper.clipEnd();
+	}
+
+	public setBonePosition(bone: string | Bone, position: IPointData): void {
+		const boneAux = bone;
+		if (typeof bone === "string") {
+			bone = this.skeleton.findBone(bone)!;
+			this.skeleton.findBone;
```
badlogic, line 307: I suppose I can remove those :)

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+			this.skeleton.findIkConstraint;
+			this.skeleton.findPathConstraint;
+			this.skeleton.findSlot;
+			this.skeleton.findTransformConstraint;
+		}
+
+		if (!bone) {
+			console.error(`Cant set bone position! Bone ${String(boneAux)} not found`);
```
badlogic, line 315: Should probably throw an error.

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+			bone.x = Spine.vectorAux.x;
+			bone.y = Spine.vectorAux.y;
+		}
+	}
+
+	public getBonePosition(bone: string | Bone, outPos?: IPointData): IPointData | undefined {
+		const boneAux = bone;
+		if (typeof bone === "string") {
+			bone = this.skeleton.findBone(bone)!;
+			this.skeleton.findBone;
```
badlogic, line 338: Remove.

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+
+		outPos.x = bone.worldX;
+		outPos.y = bone.worldY;
+		return outPos;
+	}
+
+	public static readonly skeletonCache: Record<string, SkeletonData> = Object.create(null);
+
+	public static from(skeletonAssetName: string, atlasAssetName: string, options?: ISpineOptions & { scale?: number }): Spine {
+		const cacheKey = `${skeletonAssetName}-${atlasAssetName}-${options?.scale ?? 1}`;
```
badlogic, line 362: How can cached items be evicted? Should we add a `Spine.clearSkeletonDataCache()?`

spine-ts/spine-pixi/example/assets/spineboy-polypack.atlas
```diff
@@ -0,0 +1,117 @@
+spineboy-polypack.png
```
badlogic, line 1: How can I generate a polypack file?

spine-ts/spine-pixi/example/index.html
```diff
@@ -0,0 +1,140 @@
...
+				"spineboyAtlasPolypack"
+			]);
+
+			// Create the spine display object
+			const spineBoy = spine.Spine.from("spineboySkeletonJson", "spineboyAtlasPolypack", { scale: 0.5 });
+
+			// .from(...) is a shortcut + cache for creating the skeleton data at a certain scale
+			// Here would be the "long way" of doing it (without cache):
+
+			// const skeletonAsset = Assets.get(skeletonAssetName);
```
badlogic, line 62: It's awesome that there's a high-level and low-level way of loading things.

spine-ts/spine-pixi/src/Spine.ts
```diff
@@ -0,0 +1,407 @@
...
+		this.state = new AnimationState(animData);
+		this.removeUnusedSlots = options?.removeUnusedSlots ?? false;
+		this.autoUpdate = options?.autoUpdate ?? true;
+		this.slotMeshFactory = options?.slotMeshFactory ?? ((): ISlotMesh => new SlotMesh());
+
+
+		/**
+		  * This is locked behind https://github.com/pixijs/pixijs/issues/8957
+		  * I don't want to make a custom event emitter and do `this.spineEvents.on` because that's just as "far" as `this.state.addListener`
+		  * So, until pixi fixes the custom event system, I'll stick to spine native events. - @miltoncandelero
```
badlogic, line 105: I think this is fine.

spine-ts/spine-pixi/src/assets/skeletonLoader.ts
```diff
@@ -0,0 +1,45 @@
+import type { AssetExtension, LoadAsset } from "@pixi/assets";
+import { LoaderParserPriority, checkExtension } from "@pixi/assets";
+import { ExtensionType, settings, extensions } from "@pixi/core";
+
+type SkeletonJsonAsset = any;
+type SkeletonBinaryAsset = Uint8Array;
+
+function isJson(resource: any): resource is SkeletonJsonAsset {
+	return resource.hasOwnProperty("bones");
```
badlogic, line 9: Could probably check for more fields definitely only found in a Spine skeleton .json file.

badlogic:
Alright, first code review pass is complete. It's great! Just a few questions from my end in the review. Once we have the CLA, I'll merge and polish it up. Since the public API surface won't change, my largely cosmetic changes should not affect your development.

spine-ts/spine-pixi/src/SlotMesh.ts
miltoncandelero: it is part of the vertex data but it's not during the "mesh" part of the lifecycle.
During rendering, pixi checks all objects that can be rendered together and interleaves the tint at that point.
Currently there is no way to inject an already interleaved mesh data into the data that pixi is interleaving itself.

Un-interleaving and letting pixi interleave again during the render step allows for batching between different skeleton and other pixi objects such as sprites and texts since pixi's batch renderer uploads up to 32 textures per batch

spine-ts/spine-pixi/src/SpineTexture.ts
miltoncandelero: there are 2 textures in pixi
`BaseTexture` holds the actual Image, HTMLCanvas, Compressed Texture or whatever is the actual data of your image.
`Texture` is a "view" on that BaseTexture. Mostly used to extract a rectangle or to set a pivot on a texture.

Destroying a Texture only affects that texture, but destroying a BaseTexture breaks all Textures that are a "view" of that.

Since Pixi doesn't support polygon packing and the spine runtime already handles the region extracting, I connected the Spine Texture directly to a Pixi Texture that its a view to the entire BaseTexture and let the runtime do the regioning during the Attachment processing.

Destroying `this.texture` is totally fine here. Destroying the underlaying BaseTexture (actually releasing the bytes that make up the image from memory) would mean that any other SpineTexture that depends on those same bytes will be broken too.

TLDR; 
Should "dispose" release completely the image data from memory? (if we ever need that image again we need to load the png again)

spine-ts/spine-pixi/src/SpineTexture.ts
miltoncandelero: Will ping pixi internally to see why this is the case.

spine-ts/spine-pixi/src/Spine.ts
miltoncandelero: What happens with slots that are inside skins? 🤔 

In any case, that was an idea that lead nowhere, the implementation now is not good since it removes the children one at a time. Since children is an array removing one at a time could lead to many splicings and that not the nicest way to empty an array...

I think we should remove that 😅

spine-ts/spine-pixi/src/Spine.ts
miltoncandelero: `update` is meant to be called by the user and provide the deltaTime in seconds.

`internalUpdate` is what actually updates the spine state and the autoUpdate connects directly to this. The user shouldn't call this.

The whole idea behind this is that the tick signature for pixi is something like `(deltaFrames, deltaTime)=>void` with `deltaFrames` being "how many frames passed since last frame if we were running at 60fps?". That means that at 60fps, `deltaFrame` is 1, at 30 fps `deltaFrame` is 2, at 144 fps `deltaFrame` is 2.4, etc.  
On top of that, `deltaTime` is in miliseconds.

Due to this, making a "one size fits all" update function is kinda hard. I focused on user ergonomics and going for an `update` function that the user can call with seconds (matching spine state) and internally doing the ugly math to work with the automatic update

spine-ts/spine-pixi/src/Spine.ts
miltoncandelero: oops 😅 
I was checking if you could set the position of other things besides bones

spine-ts/spine-pixi/src/Spine.ts
miltoncandelero: `skeletonCache` is public so you could call delete on each key or set them to null if you really wanted to, but we could also make a function to clear the cache 🤔

spine-ts/spine-pixi/example/assets/spineboy-polypack.atlas
miltoncandelero: image

change that to polygon during export 😅

spine-ts/spine-pixi/src/assets/skeletonLoader.ts
miltoncandelero: I guess we could try finding some metadata and maybe spine version? 🤔

miltoncandelero:
Ok, I finished the first pass of the reviews. Since the CLA is a txt how do you want it? PDF and digisign or printed, signed and scanned? 🤔 

---

on the visual bug

image

I didn't export the polygon packed spine myself, I asked somebody from discord to do it and I only took the texture and atlas file. Maybe the skeleton was not exactly the same as the one I had? 😅  
I don't do anything smartsy about the polygon packing, It's the spine runtime that creates the uvs 🤔

badlogic:
Just a PDF with a signature is enough. Sorry, I hate this too.

miltoncandelero:
Email sent and confirmation received 👍

badlogic:
Super cool, thanks! I'll get to this next week, before my holiday.

badlogic:
Uff, sorry, I didn't see your update on the visual bug. I see if I can figure it out. I'm currently merging this in an am updating the parts as discussed in the review.

badlogic:
@miltoncandelero the visual bug with polygonally packed atlases appears to be a bug in the texture packer. For Spineboy, the gun (and lower back leg part) are region attachments, meaning they will map to a rectangular area in the texture atlas page. However, due to polygonal packing, those rectangular areas in the texture atlas have pixels from other images. These show up during rendering.

We'll fix that in the editor.

badlogic:
OK, after more investigations, it seems the latest 4.1 editor release doesn't create broken atlases. @miltoncandelero could you tell me what editor version you used?

miltoncandelero:
I'm sorry but I don't know 😅

See, I don't have a spine licence so I asked somebody from the discord server to export it for me and I didn't ask what version they were using.

Sorry 😅

f0nar:
Hi guys @badlogic, @miltoncandelero.
Is this project alive? Today I tried to replace pixi-spine with this one and faced a few issues like dynamic masks (the ones with animation keys) and animated mesh deforms do not work. Is it in TODOs, issues or your minds?) @miltoncandelero have you faced these issues during testing of games that you mentioned?

In addition want to ask about purpose of this project. Probably it is more to @miltoncandelero. The `to be as out-of-the-way as possible` part is not clear for me. What does it mean?
Does this project suppose to replace pixi-spine in future?

miltoncandelero:
@f0nar I haven't had any issues with dynamic masks nor mesh deforms 🤔 

Do you have a file/minimal repro?

---

As for the second part, `pixi-spine` recreated from scratch the whole spine runtime instead of using the shared runtime that all other js engine uses. This caused it to update slowly (since changes to the official runtime had to be reverse-engineered and backported to this recreation).

By letting the official runtime do it's job, pixi code is just rendering. That is what "to be as out of the way as possible" means, this library won't try to outsmart it's parent lib (`spine-core`) and just limit itself to render whatever the runtime says it should draw 😅

f0nar:
@miltoncandelero got it. Sounds good. Then my migration plan is great idea)

Otherwise, here is a small repo with example https://github.com/f0nar/pixi-spine-testing of not working features. You can compare result to the one in `public/preview.gif` or you can run it with `pixi-spine` to see that it works there (just comment and uncomment imports in `index.ts`). Take a look at `mesh point deformation` and `animated clipping vertices`.

badlogic:
spine-pixi is a work in progress which I haven't had time to finish yet
(examples, docs, testing).

Could you please open a separate issue for the problems you ran into?

PR #2306, opened 2023-06-07
Implements AddSkin for Skin class

malikiz:
I could not find functionality in the code and documentation that would remove the added skins. It looks like this code worked.


badlogic:
This would effectively remove any skin part that is shared by skins that have been added to the custom skin. This might not be what you want.

Skins are incredibly light weight. I suggest to create a new skin if your set of combined skin changes.

PR #2329, opened 2023-07-29
Add CMake Target Exporting & option to disable testing in the build

aWilson41:
Changes:
 - Adds an option to root CMakeLists to avoid building testing.
 - Installed targets are now also exported along with config file generation.

Export Summary:
The export change is a nice "Modern CMake" update you'll find in a lot of modern libraries. This allows any code that depends on Spine to specify "SpineDIR" variable as the build directory. Then the target is automatically imported and one can link to it. For example a user could create the following demo (use this to test these changes):

main.cpp
```
#include <iostream>
#include <spine/Animation.h> // Any random include for demonstration

int main()
{
	std::cout << "Hello World!" << std::endl;
}
```
And the following build file:
CMakeLists.txt
```
cmake_minimum_required(VERSION 3.9)
project(SpineTest)

find_package(Spine REQUIRED)

add_executable(MyTest main.cpp)
target_link_libraries(MyTest PUBLIC Spine::spine-cpp) # Both include and libs setup
```

Build the above with cmake variable SpineDIR= your spine build directory

Needless Details:
configure_file is used to generate a SpineConfig.cmake in the build directory. When find_package(Spine REQUIRED) is called, a file named SpineConfig.cmake or FindSpine.cmake (old cmake) is searched for. This then includes the cmake generated spine-cpp_TARGET.cmake file when we exported earlier which will import the target into the users project. target_link_libraries to a modern cmake target then sets up the necessary includes and libraries for the MyTest target. Spine-c need not be exported as its all statically built into cpp. One might want to add export of SDL and SFML spine in the future.


badlogic:
Awesome, thanks!

Issue #2331, opened 2023-08-03
[godot] Improve performance


jordo:
We are generally finding there are a few hot spots in the spine godot runtime.  For the most part, spine-cpp is quite performant but there are a few things in the godot layer that we are hitting walls in.  I'll post some profiling runs form a CPU sampler, (both in chrome and xcode), and highlight the spots we are seeing.   I'll also be working on this today and submit a PR or two to address:

Profile from chrome.  Our scene tree `idle` takes 7.19ms, and of that 6.02ms come from spine `update_skeleton`.
<img width="1506" alt="Screen Shot 2023-08-03 at 11 20 03 AM" src="https://github.com/EsotericSoftware/spine-runtimes/assets/323868/296fec76-35db-4b2d-b849-085bd59b964b">

From that, on close inspection, a lot of time are spent here:  Mainly allocation and releasing temporary `SpineBone`s, `SpineSlot`s and also `StringName`s.   

<img width="1096" alt="Screen Shot 2023-08-03 at 11 21 51 AM" src="https://github.com/EsotericSoftware/spine-runtimes/assets/323868/450d0778-187e-4d6e-9cdf-f2bf25b19aa8">

We've encountered this a lot, and have improved a whole bunch of our godot cpp code by using static StringName's instead.  But Basically any godot API that that takes a a `const StringName&` argument ex:  `emit_signal`, you pay a huge cost by calling the function with a `const char*`.  For example:

```cpp
emit_signal("animation_started")
```   
Is actually quite costly becuase under the hood godot ends up constructing a `StringName` from the `const char*` overloaded constructor, and this ends up allocating memory and acquiring a global lock/mutex.  (Annoying I know).

At any rate, godot master and we ourselves even though it isn't backported to `3.x` use a macro `SNAME` which statically constructs a `StringName` from a `const char*` into static storage for the lifetime of the application.  This keeps the StringName around in the global map indefinitely which elimintes the construction cost of these everywhere.

`SNAME` is defined as:  
```cpp 
#define SNAME(m_arg) ([]() -> const StringName & { static StringName sname = _scs_create(m_arg); return sname; })()
```
And you can use as a replacement like:

```cpp
emit_signal(SNAME("animation_started"))
```

Line 163 seems to be quite quite heavy in our profiles...  `slot->get_bone()` end up allocating (temporarily) quite heavy godot objects, and it looks like this is called often every frame.  
[...]

PR #2332, opened 2023-08-03
Feature4.1/godot performance improvements

jordo:
Improvement addressing a bunch of perf limitations we hit wrt https://github.com/EsotericSoftware/spine-runtimes/issues/2331

A quick summary:

 -  Godot 'Objects/References' are generally quite heavy.  (post and pre initialize hooks, string allocations, map entries (registry into a global `ObjectDB`, etc.  Creating them on the fly as temporaries in hot lanes really hits hard.  

Screen Shot 2023-08-03 at 3 59 20 PM

We hit these a lot on find_bone and find_slot.   Improved a lot here by just adding a caching layer to these spots.

 - Updated many of the calls into godot functions that take a `const StringName&` to use a StringName instead of `const char*`, this is much more performant as promoting a `const char*` argument to StringName causes a StringName constructor which ends up locking a global mutex and an allocation/free cycle.   I'm not 100% up to date on whats going on in godot `master` branch, but iirc they expose the `SNAME` macro in master, it's just not backported to the `3.x` branch.

- And lastly, there's a bunch of allocations and memcopy for temporary strings in the `find*` which constructs a `spine::string` from a Godot string, but is really only used a stack temporary.  This was a little tricker to optimize as it required api support in spine's string class to take 'ownership' of a string buffer (i.e. omit the memcpy), but actually not free the buffer on destruction.  More or less a string view impl...     This isn't the prettiest solution here, as a string view class or subclass is probably bettter, but open to suggestions here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,7 +171,9 @@ spine-godot/spine_godot/spine-cpp
 spine-godot/spine_godot/__pycache__
 spine-godot/example/.import
 spine-godot/spine_godot/*.obj
+*.obj
 *.bc
+*.o
 spine-godot/example/.godot
 spine-godot/example-v4/.godot
 spine-cocos2dx/example/build-macos
--- a/spine-cpp/spine-cpp/include/spine/SpineString.h
+++ b/spine-cpp/spine-cpp/include/spine/SpineString.h
@@ -39,10 +39,11 @@
 namespace spine {
 	class SP_API String : public SpineObject {
 	public:
-		String() : _length(0), _buffer(NULL) {
+		String() : _length(0), _buffer(NULL), _tempowner(true) {
 		}
 
-		String(const char *chars, bool own = false) {
+		String(const char *chars, bool own = false, bool tofree = true) {
+			_tempowner = tofree;
 			if (!chars) {
 				_length = 0;
 				_buffer = NULL;
@@ -58,6 +59,7 @@ namespace spine {
 		}
 
 		String(const String &other) {
+			_tempowner = true;
 			if (!other._buffer) {
 				_length = 0;
 				_buffer = NULL;
@@ -82,7 +84,7 @@ namespace spine {
 
 		void own(const String &other) {
 			if (this == &other) return;
-			if (_buffer) {
+			if (_buffer && _tempowner) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 			_length = other._length;
@@ -93,7 +95,7 @@ namespace spine {
 
 		void own(const char *chars) {
 			if (_buffer == chars) return;
-			if (_buffer) {
+			if (_buffer && _tempowner) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 
@@ -113,7 +115,7 @@ namespace spine {
 
 		String &operator=(const String &other) {
 			if (this == &other) return *this;
-			if (_buffer) {
+			if (_buffer && _tempowner) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 			if (!other._buffer) {
@@ -129,7 +131,7 @@ namespace spine {
 
 		String &operator=(const char *chars) {
 			if (_buffer == chars) return *this;
-			if (_buffer) {
+			if (_buffer && _tempowner) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 			if (!chars) {
@@ -200,7 +202,7 @@ namespace spine {
 		}
 
 		~String() {
-			if (_buffer) {
+			if (_buffer && _tempowner) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 		}
@@ -208,6 +210,7 @@ namespace spine {
 	private:
 		mutable size_t _length;
 		mutable char *_buffer;
+		mutable bool _tempowner;
 	};
 }
 
--- a/spine-godot/spine_godot/SpineCommon.h
+++ b/spine-godot/spine_godot/SpineCommon.h
@@ -56,6 +56,9 @@
 #define VARIANT_FLOAT Variant::REAL
 #define GDREGISTER_CLASS(x) ClassDB::register_class<x>()
 #define GEOMETRY2D Geometry
+#ifndef SNAME
+#define SNAME(m_arg) ([]() -> const StringName & { static StringName sname = _scs_create(m_arg); return sname; })()
+#endif
 #endif
 
 #define SPINE_CHECK(obj, ret)                      \
@@ -65,6 +68,7 @@
 	}
 
 #define SPINE_STRING(x) spine::String((x).utf8())
+#define SPINE_STRING_TMP(x) spine::String((x).utf8(), true, false)
 
 // Can't do template classes with Godot's object model :(
 class SpineObjectWrapper : public REFCOUNTED {
@@ -81,9 +85,9 @@ protected:
 	void spine_objects_invalidated() {
 		spine_object = nullptr;
 #if VERSION_MAJOR > 3
-		spine_owner->disconnect("_internal_spine_objects_invalidated", callable_mp(this, &SpineObjectWrapper::spine_objects_invalidated));
+		spine_owner->disconnect(SNAME("_internal_spine_objects_invalidated"), callable_mp(this, &SpineObjectWrapper::spine_objects_invalidated));
 #else
-		spine_owner->disconnect("_internal_spine_objects_invalidated", this, "_internal_spine_objects_invalidated");
+		spine_owner->disconnect(SNAME("_internal_spine_objects_invalidated"), this, SNAME("_internal_spine_objects_invalidated"));
 #endif
 	}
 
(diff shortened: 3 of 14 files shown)


jordo:
These optimizations took our `SpineSprite::update_skeleton` cpu times from roughly ~10ms to ~4ms.  About a 2-3x improvment.  (We have many SpineSprite's on screen at once in our game).   


https://github.com/EsotericSoftware/spine-runtimes/assets/323868/ba6e821f-34fa-4ff2-9339-7195387fdda0

.gitignore
```diff
@@ -167,11 +167,12 @@ spine-cocos2dx/example/proj.android/app/.cxx
 spine-cocos2dx/example/build-win
 spine-godot/godot
 spine-godot/spine_godot/*.o
-spine-godot/spine_godot/spine-cpp
```
badlogic: Please do not remove this from the .gitignore file. The `spine-godot/build/setup.sh` script will copy `spine-cpp/` over to the appropriate place in the `spine-godot` directory.

spine-cpp/spine-cpp/include/spine/SpineString.h
```diff
@@ -200,14 +202,15 @@ namespace spine {
...
-			if (_buffer) {
+			if (_buffer && _tofree) {
 				SpineExtension::free(_buffer, __FILE__, __LINE__);
 			}
 		}
 
 	private:
 		mutable size_t _length;
 		mutable char *_buffer;
+		mutable bool _tofree;
```
badlogic: Needs a better name.

spine-godot/spine_godot/SCsub
```diff
@@ -2,10 +2,10 @@ Import('env')
 
 # Need to add the include path to env so the vsproj generator consumes it.
 if env["vsproj"]:
-    env.Append(CPPPATH=["#../spine_godot/spine-cpp/include"])
```
badlogic: All the changes in this file must be reverted.

spine-godot/spine_godot/SpineCommon.h
```diff
@@ -56,6 +56,9 @@
 #define VARIANT_FLOAT Variant::REAL
 #define GDREGISTER_CLASS(x) ClassDB::register_class<x>()
 #define GEOMETRY2D Geometry
+#ifndef SNAME
+#define SNAME(m_arg) ([]() -> const StringName & { static StringName sname = _scs_create(m_arg); return sname; })()
```
badlogic, line 60: Need to check if _scs_create is actually exposed in 3.x.

spine-godot/spine_godot/SpineSkeleton.cpp
```diff
@@ -113,26 +113,34 @@ void SpineSkeleton::set_slots_to_setup_pose() {
 Ref<SpineBone> SpineSkeleton::find_bone(const String &name) {
 	SPINE_CHECK(skeleton, nullptr)
 	if (EMPTY(name)) return nullptr;
-	auto bone = skeleton->findBone(SPINE_STRING(name));
+	auto bone = skeleton->findBone(SPINE_STRING_TMP(name));
```
badlogic, line 116: How often per frame do you call this for this change to make a difference performance wise? It would be great to have a test scene that's close to your real-world scenario, where we can measure any improvements. Profilers don't always tell the truth.

spine-godot/spine_godot/SpineSkeleton.h
```diff
@@ -66,6 +68,9 @@ class SpineSkeleton : public REFCOUNTED {
 	spine::Vector<float> bounds_vertex_buffer;
 	Ref<SpineSkin> last_skin;
 
+	std::unordered_map<spine::Bone*, Ref<SpineBone>> _cached_bones;
```
badlogic, line 71: When the `SpineSkeleton` is destroyed, will the `Ref`s to the bones and slots be decreased as well?

spine-godot/spine_godot/SpineSprite.cpp
```diff
@@ -665,15 +665,15 @@ void SpineSprite::update_skeleton(float delta) {
 		!animation_state->get_spine_object())
 		return;
 
-	emit_signal("before_animation_state_update", this);
+	emit_signal(SNAME("before_animation_state_update"), this);
```
badlogic, line 668: These `SNAME`s here are the only ones that will possibily make a difference in terms of #allocations, as they are created each frame. The rest won't hurt, but they won't make any difference performance wise.

spine-godot/spine_godot/spine-cpp
```diff
@@ -0,0 +1 @@
+../../spine-cpp/spine-cpp
```
badlogic: I'm reviewing on GitHub, but this looks like a symlink? Can't have that I'm afraid.

badlogic:
Could you quantify "many sprites"? Because 5 is definitely not many.

In the spine-godot project, there's a test scene that tests batching. It draws around 100 Spineboys, which are quite heavy compared to what you show in the video, and it hits way below 16ms per frame on mid-range desktop hardware with a debug build.

I went through the changes and they look good overall. Caching of slot and bone wrappers around `spine::Slot` and `spine::Bone` makes sense, as does `SNAME` in the update loop and `SPINE_STRING_TMP` when talking to the Spine API via strings.

The change to the string class is indeed a bit unfortunate, but it makes sense in this context and is a non-destructive change.

I've added a view comments for you to review. If we merge this, I'll need a signed CLA from you, see CONTRIBUTING at the end of the README.md file:
https://github.com/EsotericSoftware/spine-runtimes#contributing

Sorry for the red tape. Just send us a digital copy is good enough.

.gitignore
jordo: Will do, I'll go through the PR and clean a bunch of the other commits up or (remove completely).

spine-cpp/spine-cpp/include/spine/SpineString.h
jordo: Ya I was not actually not 100% sure on this...  I don't love the workaround, but we definitely need a way to interact with the spine api (which takes `spinestring`) without doing an additional alloc and memcopy for a stack temporary.   

Perhaps a better design would be to leave this class implementation alone and write a `SpineStringView` class which inherits `SpineString` (so we can pass them into the spine api), but the SpineStringView would have nothing to do with the backing character memory it's initialized with...  I don't know.

spine-godot/spine_godot/SCsub
jordo: Will do.

spine-godot/spine_godot/SpineCommon.h
jordo: I think it is???  I 'll have to check, we are using a custom 3.x fork...

spine-godot/spine_godot/SpineSkeleton.cpp
jordo: `SpineSkeleton::find_bone` wasn't called by us manually, but the function below it `SpineSkeleton::find_slot` is called internally in `SpineSlotNode::update_transform` which gets hit very often.     I just though applying the same optimization to  `find_bone` as `find_slot` was just kind defensive performance coding as I could see it could be source of performance issues as well.

spine-godot/spine_godot/SpineSkeleton.h
jordo: Yes should be.  I was unsure about the map here though.  I didn't see any other stdlib dependencies in godot_spine so I wasn't sure what you prefer for convention here.  I also wasn't sure what the pointer stability of `spine::Bone*` is either.  Can the pointer be updated to point at a different bone, or is it static for the lifetime of the skeleton?   Another option is to have the string as the map key, but my thought was the pointer is the faster map lookup.

spine-godot/spine_godot/SpineSprite.cpp
jordo: YA, some definitely not needed, but I just opted for consistency everywhere in spine_godot .  My convention is to always use SNAME for any `const char*`  godot API that takes `StringName`.

spine-godot/spine_godot/spine-cpp
jordo: Will remove (sorry artifact from how we setup our projects)

jordo:
> Could you quantify "many sprites"? Because 5 is definitely not many.

Ya sorry, maybe 20-30 SpineSprite's...  I just attached a quick screencap but there are other sprites in the frame processing.

> Sorry for the red tape. Just send us a digital copy is good enough.

I'll cleanup a bunch of this, but wanted to post the PR to get your initial review before cleaning up, polishing, and potentially refactoring anything...  I think I've commited a few PR's already so maybe I have signed the CLA already?  (I can't remember).

jordo:
For reference, a quick view of our actual scene setup.  We have 8 SpineSlotNodes per SpineSprite, and about 30 SpineSprites on those profiling runs.

image

spine-godot/spine_godot/spine-cpp
jordo: Oh, looking back I guess we did this so we could setup compilation of the module without an additional 'copy to build' step.   As the symlink is setup to spine-cpp location in the github repo.  Will still remove out of this PR though.

jordo:
converted to draft as I cleanup.  Will rebase on final PR.

jordo:
OK, i think this is ready for review.

spine-godot/spine_godot/SpineSprite.cpp
badlogic: Makes sense!

badlogic (approved):
Looks good! I'll merge this today and test it across Godot versions.

Thanks, really appreciate it! Please don't forget about the CLA.

badlogic:
NM the CLA comment, looks like we got it!

badlogic:
Sorry, took a bit! Merged and tested. Thank you!

PR #2337, opened 2023-08-08
[Phaser] Adds Asset Pack support

PhaserEditor2D:
Phaser provides the Asset Pack format for defining the assets in a manifest (JSON) file.

This pull request modifies the `SpineSkeletonDataFile` and `SpineAtlasFile` classes for implementing the Asset Pack feature. The modifications are about extracting the file info from the `key` argument of the constructor. The Phaser loader uses the file implementations for processing the asset pack entries. It passes the JSON info as the `key` parameter of the file constructor. The modifications in this PR follow the same pattern in the implementation of other Phaser files.

Here I attach a working demo. The `spine-phaser.js` runtime contains the modifications of this PR.

[demo-spine.zip](https://github.com/EsotericSoftware/spine-runtimes/files/12287323/demo-spine.zip)

### DIFF
--- a/spine-ts/spine-phaser/src/SpinePlugin.ts
+++ b/spine-ts/spine-phaser/src/SpinePlugin.ts
@@ -259,13 +259,13 @@ class SpineSkeletonDataFile extends Phaser.Loader.MultiFile {
 	constructor (loader: Phaser.Loader.LoaderPlugin, key: string, url: string, public fileType: SpineSkeletonDataFileType, xhrSettings: Phaser.Types.Loader.XHRSettingsObject) {
 		
 		if (Phaser.Utils.Objects.IsPlainObject(key as any)) {
-            const config = key as any;
-            key = Phaser.Utils.Objects.GetFastValue(config, "key");
-            url = Phaser.Utils.Objects.GetFastValue(config, "url");
+            		const config = key as any;
+            		key = Phaser.Utils.Objects.GetFastValue(config, "key");
+            		url = Phaser.Utils.Objects.GetFastValue(config, "url");
 			const type = Phaser.Utils.Objects.GetFastValue(config, "type");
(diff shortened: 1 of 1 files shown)


spine-ts/spine-phaser/src/SpinePlugin.ts
```diff
@@ -257,6 +257,16 @@ enum SpineSkeletonDataFileType {
 
 class SpineSkeletonDataFile extends Phaser.Loader.MultiFile {
 	constructor (loader: Phaser.Loader.LoaderPlugin, key: string, url: string, public fileType: SpineSkeletonDataFileType, xhrSettings: Phaser.Types.Loader.XHRSettingsObject) {
+		
+		if (Phaser.Utils.Objects.IsPlainObject(key as any)) {
```
badlogic: This cast validates the `key` parameters type, which is `string`. Can you provide an interface definition for the object?

badlogic (changes requested):

spine-ts/spine-phaser/src/SpinePlugin.ts
PhaserEditor2D: Ok, I did it.

badlogic:
Sorry for the delay, good stuff!

PhaserEditor2D:
Nice, thanks!

stevium:
@PhaserEditor2D  I think there is an issue with this PR in case `BaseUrl` is set. The `atlas` and `json` are resolved correctly, but the `.png` files get the `BaseUrl` prepended to the `url` which already contains the `BaseUrl`. So for example if base url is `src/assets` the end `src` gets resolved to `src/assets/src/assets/<ATLAS_URL_FROM_PACK.JSON/fileName.png>`. It only works in case I remove `BaseUrl` or set it to an empty string.

Issue #2342, opened 2023-08-12
[Phaser] SCREEN blendmode not working


spayton:
The SCREEN blendmode does not work.  This hasn't worked since the 3.8 ts runtimes.

Here's my test case, a multicoloured solid filled circle drawn over a split gradient background, in each or the 4 blend modes (normal, additive,multiply,screen).

This is captured from the spine editor v4.1.23:

![image](https://github.com/EsotericSoftware/spine-runtimes/assets/4949575/514483b2-bca9-4b9f-8622-f6087ae36851)

This is captured from a Phaser 3.60 + official spine plugin 4.1

![image](https://github.com/EsotericSoftware/spine-runtimes/assets/4949575/2ca12039-3cf8-4914-93c4-b4e16385380a)

Only the NORMAL blend should show a solid circle.  Note the difference between the SCREEN versions, the spine editor is correct, whilst the Phaser version is solid and the same as NORMAL.

I have tried both with and without PMA, it makes no difference, the Phaser version is wrong and in the same way wrong.

I also tried Pixi 7.2.4 with its own spine player, this got the SCREEN blend right, as follows.

![image](https://github.com/EsotericSoftware/spine-runtimes/assets/4949575/fb071565-295a-4f8e-8311-3d7a7fc2cc46)

Having previously compared the Phaser & Pixi rendering, the difference was down to having to single or double blend functions, which I tried to bodge that behaviour into the spine-ts runtime but never quite got it working before leaving it.  Now I have a renewed interest in getting this working and now we have an official Phaser plugin.

Looking through the PR's I notice this [https://github.com/EsotericSoftware/spine-runtimes/pull/2025] for spine-ts which seems to support my earlier findings, though I wonder why it's been hanging there for so long.

spayton:
Right, I now have a working version.

The destination blend needs splitting into color and alpha, so modify all the code using blendmodes to use an extra value.

At this point my change is very similar the the PR mentioned above, but differs in two ways:

1. I've made changes to more files and still need to check the examples.
2. My WebGLBlendModeConverter is different, with more differentiation between the PMA and non-PMA modes.

Testing the Phaser plugin now shows all blendmodes in both PMA/non-PMA rendering the same as the spine editor and skeletion viewer.  When I think I've covered everything I'll PR it.

PR #2347, opened 2023-08-22
[TS] Fix Webgl Blendmodes

spayton:
This fixes the blendmode problem mentioned in https://github.com/EsotericSoftware/spine-runtimes/issues/2342#issue-1848207378

The fix is to split webgl blendmodes into separate destination color/alpha values, instead of a common destination blend value which cannot implement SCREEN blending.

This handles both PMA and non-PMA.

### DIFF
--- a/spine-ts/spine-webgl/src/LoadingScreen.ts
+++ b/spine-ts/spine-webgl/src/LoadingScreen.ts
@@ -88,7 +88,7 @@ export class LoadingScreen implements Disposable {
 
 		renderer.resize(ResizeMode.Expand);
 		renderer.camera.position.set(canvas.width / 2, canvas.height / 2, 0);
-		renderer.batcher.setBlendMode(gl.ONE, gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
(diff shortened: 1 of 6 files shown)


badlogic:
Fantastic work, thank you! This has been a headscratcher for me for a while. The other PR didn't entirely fix it and I never found the time to wokr it out.

badlogic:
@spayton whoopsies, this broke premultiplied alpha rendering.
Screenshot 2023-09-13 at 19 32 48

spayton:
Well, not surprised something somewhere bust, but I'm surprised I heven't seen this as I do it a lot.  Is that one of the examples?

badlogic:
Jupp. Working on it. I'll update you here.

badlogic:
This is now fixed in the latest 4.1 and 4.2-beta branch commits as well as the latest 4.1 and 4.2 NPM packages. I rewrote how blend modes are handled entirely, it's now quite a bit cleaner.

spayton:
Just tested, at a glance the blendmodes seems fine.

badlogic:
I've rewritten the blending, works as intended now in all cases.

PR #2350, opened 2023-08-24
[cocos creator] the first cocos creator spine code.

Canvasfull:
The first cocos creator spine code submission.


badlogic:
As per our email exchange, we'll wait until Cocos has finished their plugin architecture. Then we can provide plugins for all supported Spine Runtimes versions, instead of just one.

Issue #2333, opened 2023-08-04
[Phaser] plugin object name conflict with existing plugin


spayton:
This is more of a suggestion than notification of a problem.

This new offical spine plugin for the Phaser3 framework is great and a highly welcome addition.  In use I have found one aspect of it that could be more flexible, that would have multiple benefits.

This is the name of the spine object type, currently it is 'spine' the same as Phaser's own spine plugins.  This is the property name that is used to generate the functionality such as

```js
scene.add.spine
```

This constant is set in src/keys.ts

```js
export const SPINE_GAME_OBJECT_TYPE = "spine";
```

In practice, this can cause confusion because this has been used up till now solely by Phaser's own spine plugins which have a different API.  All of Phaser's current spine users will be used to using that game object type with its own API.

I'm suggesting that this game object type can be changed from the default 'spine' by the user.  Being able to do this would have the following benefits:

- there would be no confusion over which spine object was being created, thus no confusion over which API is being used eg.
```js
scene.add.spine    // legacy plugin
scene.add.spinePhaser    // official plugin
```
- multiple spine plugins could be loaded and be operational, ie phasers own spine plugin could co-exist with the new official one. This is quite a benefit, it would allow:
  - easy project conversion to official plugin by tackling spine asset usage one at a time, rather than the 'all in' approach you would get by swapping to a different plugin and everyhthing being broken at the same time.
  - the ability to be able to use older spines with newer versioned ones, ie wouldn't have to update all spines to same version just to suit one plugin version

I'm only suggesting adding functionality.  The default behaviour would remain unchanged, and would suit Phaser users with no previous spine usage or spine baggage.

I can easily achieve this by editing the SPINE_GAME_OBJECT_TYPE string in the plugin src and rebuilding, but it seems such a useful addition that it should be available 'out-of-the-box'.

badlogic:
It's an interesting idea, though I'm of the opinion that mixing plugins and versions is a very bad thing. It leads to version conflicts, confusion which asset files are compatible with what, etc. It is a one time thing to update, it's a continuous pain to keep multiple versions around. I'd not officially endorse this.

However, making it configurable is still sensible. The problem is: I don't know how. There's no mechanism to inject something into a plugin before it is registered, at least not as far as I can see. Any ideas?

spayton:
[...]

PR #2357, opened 2023-09-05
[Phaser] Fix GameObject input not working

spayton:
Enabling the input on a spine GameObject with setInteractive() added it to the wrong scene input list, so was never hit.

The new SpineGameObject should use the gameObject's own scene ref, not the system scene.

### DIFF
--- a/spine-ts/spine-phaser/src/SpinePlugin.ts
+++ b/spine-ts/spine-phaser/src/SpinePlugin.ts
@@ -124,7 +124,7 @@ export class SpinePlugin extends Phaser.Plugins.ScenePlugin {
 
 		let self = this;
 		let addSpineGameObject = function (this: Phaser.GameObjects.GameObjectFactory, x: number, y: number, dataKey: string, atlasKey: string, boundsProvider: SpineGameObjectBoundsProvider) {
-			let gameObject = new SpineGameObject(scene, self, x, y, dataKey, atlasKey, boundsProvider);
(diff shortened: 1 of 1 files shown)


badlogic:
That seems to be a 3.60.0 peculiarity, no? I had a related issue today that worked fine in 3.55.2 but failed in 3.60.0 because that seems to have introduced the system scene. I'm now paranoid that there are other places where things are messed up due to that.

Thanks for the fix!

badlogic:
I updated all examples to 3.60.0, tested things, and published new 4.1 and 4.2 NPM packages. Thanks again for your continued help with ironing out spine-phaser issues!

spayton:
Unfortunately I connot speak with reliable authority about the system scene or the global/scene plugin usage, maybe PhotonStorm could cast an eye over this, to ensure it's behaving as intended.  But yes, I wouldn't be surprised if there's some more problems or inneficiencies hiding.

I'm just tracking down issues as they arise, this one just looked like a typo, which I checked against behaviour of Phaser's own spine plugin.  The new functionality added here #2333 allowed me to quickly have both plugins running side by side again for comparison 😃.

It's possible this wasn't a problem on older Phaser versions, but I didn't check, there's too many compatibility problems to make it easy enough to try in a mature app.  There are not only breaking changes between 3.55.x/3.60, but also within the 3.60 beta's.  3.60+ is the version to be working with.

PR #2366, opened 2023-09-15
Package examples as samples

rthery:
I had warnings with .meta of materials in the Examples folder being modified in the Editor while the module was used as a package. 

```
The package cache was invalidated and rebuilt because the following immutable asset(s) were unexpectedly altered:
Packages/com.esotericsoftware.spine.urp-shaders/Examples/3D/Spine Skeletons/RaptorURP/raptor_Material.mat
```

Best practice for Unity packages is to distribute examples as samples, cf. https://docs.unity3d.com/Manual/cus-samples.html

I've forked to fix it, but feel free to either merge or decline if you want to have it upstream (I'd be of course very happy to see this fixed upstream!)


HaraldCsaszar:
Thanks very much for the PR! We've just pushed a slightly different version in the above commit.

HaraldCsaszar:
A new 4.1 Spine URP Shaders UPM package is available for download:

PR #2395, opened 2023-10-06
[unity] Set fps at generated SkeletonMecanim animation clip assets

NamYCM:
In some case, the FPS of that animation clip is very important to detect which frame the animation is running at. That's why the we need to sync the FPS of spine animation and generated animation clip


HaraldCsaszar:
@NamYCM I'm very sorry about the very late reply, I have completely missed your issue ticket! We've integrated the proposed changes on the 4.2 branch. A new spine-unit 4.2 unitypackage is available for download:

PR #2410, opened 2023-11-03
[pixi][phaser][threejs] Fixed "ReferenceError: window is not defined" in Node.js environment. See #2409

### LINKED ISSUE #2409: [pixi] ReferenceError: window is not defined in Node.js
**Problem:**
https://github.com/EsotericSoftware/spine-runtimes/blob/fa4310b4d3f44f5b5c9cad050f46dfe19f1bb831/spine-ts/spine-pixi/src/require-shim.ts#L35

In `require-shim.ts` we have this code:
```
if (window.PIXI) {
	let prevRequire = window.require;
	window.require = (x: string) => {
		if (prevRequire) return prevRequire(x);
		else if (x.startsWith("@pixi/")) return window.PIXI;
	}
}
```

Because `window` doesn't exist in Node.js we have problem in Node.js when we try to use testing libraries like vitest. 

We've got this error:
```
ReferenceError: window is not defined
 ❯ ../node_modules/@esotericsoftware/spine-pixi/src/require-shim.ts:35:1
 ❯ ../node_modules/@esotericsoftware/spine-pixi/dist/index.js:3:31
```
image

**Solution:**
Add `typeof window !== 'undefined'`:
```
if (typeof window !== 'undefined' && window.PIXI) {
	let prevRequire = window.require;
	window.require = (x: string) => {
		if (prevRequire) return prevRequire(x);
		else if (x.startsWith("@pixi/")) return window.PIXI;
	}
}
```

vadimgoncharov:
Fixes https://github.com/EsotericSoftware/spine-runtimes/issues/2409

### DIFF
--- a/spine-ts/spine-phaser/src/require-shim.ts
+++ b/spine-ts/spine-phaser/src/require-shim.ts
@@ -30,7 +30,7 @@
 declare global {
 	var require: any;
 }
-if (window.Phaser) {
+if (typeof window !== 'undefined' && window.Phaser) {
 	let prevRequire = window.require;
 	window.require = (x: string) => {
 		if (prevRequire) return prevRequire(x);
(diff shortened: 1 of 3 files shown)


badlogic:
Thank you!

PR #2417, opened 2023-11-14
[unity] Explicitly specify 'SpineAtlasAsset' as a variable named 'atlasAsset'

FreeGoStudio:
…asAsset'.

Explicitly specify 'SpineAtlasAsset' as a variable named 'atlasAsset'. Reduce the error 'Must have AtlasAsset variable!' when creating a variable with a name other than 'AtlasAsset' 'but somebody don't know why.


HaraldCsaszar:
@FreeGoStudio Sorry for the late reply, I've overlooked your pull request!

I'm not sure how you encountered the issue you are describing, or how the pull request would resolve the issue, since the following line in `SpineAttributeDrawers.cs` should also have the same effect already:
https://github.com/EsotericSoftware/spine-runtimes/blob/4.2-beta/spine-unity/Assets/Spine/Editor/spine-unity/Editor/SpineAttributeDrawers.cs#L614
Could you please describe how we can reproduce the issue you have encountered?

PR #2436, opened 2023-12-19
[csharp] Exposed list optimizations

alexfrosts:
This commit handles unnecessary work with array.

Making unused items of array to default makes us iterate other all array every time, this slows work of our application dramatically. I believe, that you should do this somehow the other way. 
At least, if you want to ensure handling refs, check if your item is valueType, so you would not iterate other all array trying to set defaults which slower CPU. Anyway I believe that only when array is destroyed, working with its items stops in your app.

Also, you should compare sizes not with items.Length, but with Capacity.

I believe that you could handle working with your exposed list more accurately in individual cases.

By the way we work with il2cpp in our app and this leads to some slowing CPU cases while working with your arrays.
if you build app with il2cpp, there would be various code generations for null-checks, array boundary checks and etc during working with arrays. You could see [Il2CppSetOptionAttribute in unity docs](https://docs.unity3d.com/Manual/IL2CPP.html). This leads us to huge slowing of CPU work during iterations other array and etc. If you are sure you are working with arrays correctly, in what I believe (check your algorithms), you could turn off this code generation in il2cpp. This made work with our app faster up to 4-6 times. Especially during working with clipping zones and animations.

```
[Il2CppSetOption(Option.NullChecks, false), Il2CppSetOption(Option.ArrayBoundsChecks, false)]
CurveTimeline.GetCurveValue
```

```
[Il2CppSetOption(Option.NullChecks, false), Il2CppSetOption(Option.ArrayBoundsChecks, false)]
SkeletonCLipping.Clip (float x1, float y1, float x2, float y2, float x3, float y3, ExposedList<float> clippingArea, ExposedList<float> output)
```
I am sure that you could find much more cases there you could make your Spine Runtime much faster.
Just check work with arrays and il2cpp build cases.

### DIFF
--- a/spine-csharp/src/ExposedList.cs
+++ b/spine-csharp/src/ExposedList.cs
@@ -76,7 +76,7 @@ namespace Spine {
 		public void Add (T item) {
 			// If we check to see if we need to grow before trying to grow
 			// we can speed things up by 25%
-			if (Count == Items.Length)
+			if (Count == Capacity)
 				GrowIfNeeded(1);
 			Items[Count++] = item;
 			version++;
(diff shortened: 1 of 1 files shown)


HaraldCsaszar:
@alexfrosts Sorry for the late reply. I've overlooked this pull request for a long time (or perhaps didn't understand it when reading it the first time and then forgot about it).

Regarding your proposed improvements:

> Making unused items of array to default makes us iterate other all array every time, this slows work of our application dramatically. I believe, that you should do this somehow the other way

I assume you mean the `Resize()` call. This makes sense indeed, we will have a look at each call and improve it accordingly where suitable.

> Also, you should compare sizes not with items.Length, but with Capacity.

This makes no sense, `Capacity` is just a property getter wrapping `Items.Length`, so this would be detrimental to de-inline.

> Il2CppSetOption(Option.ArrayBoundsChecks, false)]

Thanks for the hint, much appreciated. I incorrectly assumed that this can be set globally, but it appears it can only be set per class or method call. We will see what we can do.

HaraldCsaszar:
Resolved differently via above commit on the 4.3-beta branch. Thanks for reporting and sorry for the long delay!

A new spine-unity 4.3-beta unitypackage is available:

PR #2437, opened 2023-12-22
Update README.md

royalstream:
Small fix in the documentation, it was mentioning spine-godot when it should mention spine-flutter.


badlogic:
Whoopsies and cheers!

PR #2439, opened 2024-01-05
[ts][pixi] Fix not working alpha channel in mesh animations

vadimgoncharov:
Alpha channel is not working in pixi animations right now. I prepared demo with rotating and fading object, so we can see that rotating animation is working, but fading animation is not:
alpha-not-working mov
As you can see we have changing alpha in auxColor[3], but it has no effect on scene.

The problem is in SlotMesh's `tint`:
https://github.com/EsotericSoftware/spine-runtimes/blob/73318d6c79b34ca8201e91e120e85a3cc2f9dacd/spine-ts/spine-pixi/src/SlotMesh.ts#L96-L101

[Tint](https://pixijs.download/dev/docs/PIXI.Mesh.html#tint) in Pixi (as I understand) is not working with [alpha](https://pixijs.download/dev/docs/PIXI.Mesh.html#alpha). If we want to use alpha, we should directly set it:
```
SlotMesh.auxColor[0] = finalVertices[2];
SlotMesh.auxColor[1] = finalVertices[3];
SlotMesh.auxColor[2] = finalVertices[4];
SlotMesh.auxColor[3] = finalVertices[5];

this.tint = SlotMesh.auxColor;
this.alpha = SlotMesh.auxColor[3]; // <--- here
```

When we do this, everything starts to work :)
alpha-working mov

### DIFF
--- a/spine-ts/spine-pixi/src/SlotMesh.ts
+++ b/spine-ts/spine-pixi/src/SlotMesh.ts
@@ -99,6 +99,7 @@ export class SlotMesh extends Mesh implements ISlotMesh {
 		SlotMesh.auxColor[3] = finalVertices[5];
 
 		this.tint = SlotMesh.auxColor;
+		this.alpha = SlotMesh.auxColor[3];
 		this.blendMode = SpineTexture.toPixiBlending(slotBlendMode);
 
 		if (this.geometry.indexBuffer.data.length !== finalIndices.length) {


vadimgoncharov:
There is same problem in DarkSlotMesh: https://github.com/EsotericSoftware/spine-runtimes/blob/73318d6c79b34ca8201e91e120e85a3cc2f9dacd/spine-ts/spine-pixi/src/DarkSlotMesh.ts#L81-L100

But I don't know difference between `this.darkTint` and `this.tint`. Should we also use `this.alpha = ...` here?

badlogic:
Oh, that's great! I was already looking into this and couldn't figure it out. Thank you!

PR #2441, opened 2024-01-07
[spine-ts][core] Fix when physics is none throw error

Jerry457:


badlogic:
Ugh, good catch, thank you! JavaScript and falsy-ness does my head in.

Jerry457:
> Ugh, good catch, thank you! JavaScript and falsy-ness does my head in.
Have you fixed it?

PR #2454, opened 2024-01-22
[flutter] Fix skeleton.getY() and skeleton.getScaleY()

k41531:
I have fixed a bug where the function for obtaining the Y position and size from the Skeleton was incorrectly returning X.


badlogic:
Ugh, good catch. Thanks!

PR #2458, opened 2024-02-19
[flutter] Fix missing definition of Blend mode set function

k41531:
Fixed a function on the C++ side that was not defined that is called by Flutter.


badlogic:
Ouch, good catch!

PR #2470, opened 2024-03-05
flutter: fix memory cleanup is slow due to not destroying shaders

hantrungkien:


badlogic:
Good catch, thanks!

hantrungkien:
@badlogic What do I need to add so it can be merged?

badlogic:
Nothing, I pressed the wrong button :/

Merged!

hantrungkien:
@badlogic thanks. When can I get 4.2.3 from pub.dev?

badlogic:
4.2 is currently unstable as we are wrapping up the 4.2 release. I'll publish a new 4.1 tomorrow.

hantrungkien:
@badlogic please release help me

badlogic:
Your fix is incorrect and will result in crashes, as the same image is disposed multiple times with this approach.

badlogic:
I've reverted the fix and instead resorted to cleaning out the list of paints. `ImageShader` instances are not supposed to be disposed manually according to the docs. This should have the same effect once the GC kicks in.

Issue #2346, opened 2023-08-21
[unity] ArgumentOutOfRangeException at SkeletonGraphic.PrepareRendererGameObjects


HaraldCsaszar:
Reported on the forum here:
https://esotericsoftware.com/forum/d/24350-skeletongraphic-with-separatorparts-not-scaling-correctly/7

sandolkakos:
Hi @HaraldCsaszar, since we needed to update the spine version to make use of the on-demand module, we started seeing that issue again. But it is hard to know what is the reproduction for it since we have only this Crashlytics data:

---
<img width="700" alt="image" src="https://github.com/EsotericSoftware/spine-runtimes/assets/1542861/90e31cd9-1810-4344-afe2-7401d00240e1">

<img width="246" alt="image" src="https://github.com/EsotericSoftware/spine-runtimes/assets/1542861/b060ff1f-bdd7-4c15-8efa-be9284d7d53a">

HaraldCsaszar:
Closing the issue as we could not yet reproduce the issue and received no further report. Might be no longer an issue on 4.3-beta. We're happy to reopen the issue if the issue re-occurs with a reproduction project.

PR #2478, opened 2024-03-08
[spine-unity] Fix for the issue [2346] causing canvasRenderers and submeshGraphics to get unsynced

sandolkakos:
Fix for the [issue 2346](https://github.com/EsotericSoftware/spine-runtimes/issues/2346) causing canvasRenderers and submeshGraphics to get unsynced and throwing null exception reference at:

https://github.com/EsotericSoftware/spine-runtimes/blob/3b8069f4b91cc05329147c372446cd5b54e0f5dd/spine-unity/Assets/Spine/Runtime/spine-unity/Components/SkeletonGraphic.cs#L1055

### DIFF
--- a/spine-unity/Assets/Spine/Runtime/spine-unity/Components/SkeletonGraphic.cs
+++ b/spine-unity/Assets/Spine/Runtime/spine-unity/Components/SkeletonGraphic.cs
@@ -991,6 +991,20 @@ namespace Spine.Unity {
 #if UNITY_EDITOR
 			RemoveNullCanvasRenderers();
 #endif
+			foreach (var currentCanvasRenderer in canvasRenderers) {
+				if (!currentCanvasRenderer.TryGetComponent<SkeletonSubmeshGraphic>(out var submeshGraphic)) {
+					AddNewSubmeshGraphic(currentCanvasRenderer.gameObject);
+					continue;
+				}
+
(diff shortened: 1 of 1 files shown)


HaraldCsaszar:
@sandolkakos Thanks for the pull request! Unfortunately we have to decline it this time since `SyncSubmeshGraphicsWithCanvasRenderers` is not intended to be called every frame, as we want to avoid calling `canvasRenderer.GetComponent<SkeletonSubmeshGraphic>();` that frequently.

It would be great to know why canvasRenderers and submeshGraphics can get out of sync, then we could provide a more cost-efficient bugfix.

sandolkakos:
> @sandolkakos Thanks for the pull request! Unfortunately we have to decline it this time since `SyncSubmeshGraphicsWithCanvasRenderers` is not intended to be called every frame, as we want to avoid calling `canvasRenderer.GetComponent<SkeletonSubmeshGraphic>();` that frequently.
> 
> It would be great to know why canvasRenderers and submeshGraphics can get out of sync, then we could provide a more cost-efficient bugfix.

Ohh I see. Fair enough, I haven't checked it was called from Update! 😅
I will try to find a time on the next weeks to see if I can find the reason for both arrays getting out of sync in our project.

HaraldCsaszar:
@sandolkakos Thanks very much for your understanding and for all your efforts! :)

PR #2491, opened 2024-03-26
[spine-c] Fix crashes in skins or animations error handling

DimonSE:
Good day,

I've implemented additional fixes related to pull request #2276, aimed at correctly unloading the spine skeleton when encountering skin or animation errors.

In most cases, everything works well as is, but sometimes crashes occurred because, in case of some problems, it is necessary to call spSkeletonData_dispose(), which tries to clear all skins and animations, even those that have not fully loaded or were not loaded at all.

In SkeletonJson, the fix is quite simple; I just adopted the approach used in the /* Bones. */ section, where skeletonData->bonesCount++; is only called at the end, instead of immediately setting skeletonData->bonesCount = bones->size;
This allows spSkeletonData_dispose() to clear only those bones that were actually created, rather than attempting to delete all bones. I simply repeated this approach for slots, various constraints, and animations. 
Thus, no matter at what stage of loading inside spSkeletonJson_readSkeletonData() an exit from the function is made, skeletonData will be correctly deleted without crashes when an error happened, and I will get a valid string in pSkelJson->error; with a detailed description of what exactly is wrong.

SkeletonBinary.c required simpler adjustments, mainly reducing error checks to skins and animations.
I've limited the deletion scope by setting skeletonData->skinsCount = i + 1; and skeletonData->animationsCount = i + 1; in their respective error handlers.

Additionally, service function calls were added to minimize memory leaks during erroneous skeleton processing.

These enhancements help our team of animators and designers to independently obtain more information when checking animations, specifically which skin or animation was broken, allowing them to more frequently identify problems themselves by using our internal 'assets viewer' to check .skel files, without involving game programmers to debug the issue.


badlogic:
Cheers!

PR #2495, opened 2024-04-03
Dont do init transform if you dont need to.

jasonwinterpixel:
`init_transform()` is expensive!

This PR simply doesnt do it if the bone mode is a no op!


badlogic:
Cheers! I just wonder, in what case would this have any significant impact on a game's performance?

jasonwinterpixel:
We're running our games on mobile web. So when compiled to WASM and running in Chrome on some old android device, I'll take every fraction of a millisecond I can get.

PR #2523, opened 2024-05-10
support premultiplied alpha

oldeviloldevil:
these may support premultiplied alpha


badlogic:
Cheers! I've cleaned it up and ported it to the C backend as well.

PR #2540, opened 2024-06-04
Add IsNextReady to TrackEntry

TyelorD:
This allows users to check when/if the next TrackEntry in a sequence of TrackEntries is applied (i.e. TrackEntries added to the currently playing TrackEntry using AnimationState.AddAnimation.

This is intended to avoid exposing any additional internal members that could be confusing to some users.

This is to avoid the issue that when TrackEntry has a non-null next member, using TrackEntry.Next.WasApplied will always return true, regardless of whether it has ever actually been applied.

If any changes are needed, or this logic doesn't work as it seems to with my use-case, or if these kinds of pull-requests are unsuitable, then please let me know. Thank you for this awesome tool!

### DIFF
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -1217,7 +1217,7 @@ namespace Spine {
 		/// <summary>Returns true if the next track entry is about to be applied on the next update call.</summary>
 		/// <seealso cref="AnimationState.Apply(Skeleton)"/>
 		public bool WillNextBeAppliedOnUpdate {
-            get { return next != null && System.MathF.Abs(nextTrackLast - next.delay) >= float.Epsilon; }
+            get { return next != null && System.MathF.Abs(nextTrackLast - next.delay) < float.Epsilon; }
         }
 
 		/// <summary>
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -1216,8 +1216,8 @@ namespace Spine {
 
 		/// <summary>Returns true if the next track entry is about to be applied on the next update call.</summary>
 		/// <seealso cref="AnimationState.Apply(Skeleton)"/>
-		public bool WillNextBeAppliedOnUpdate {
-            get { return next != null && System.MathF.Abs(nextTrackLast - next.delay) < float.Epsilon; }
+        public bool IsNextReady {
+			get { return next != null && nextTrackLast - next.delay >= 0; }
         }
 
 		/// <summary>
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -1214,9 +1214,9 @@ namespace Spine {
 			get { return nextTrackLast != -1; }
 		}
 
-		/// <summary>Returns true if the next track entry has been applied at least once.</summary>
+		/// <summary>Returns true if the next track entry is about to be applied on the next update call.</summary>
 		/// <seealso cref="AnimationState.Apply(Skeleton)"/>
-		public bool WasNextApplied {
+		public bool WillNextBeAppliedOnUpdate {
             get { return next != null && System.MathF.Abs(nextTrackLast - next.delay) >= float.Epsilon; }
         }
 
--- a/spine-csharp/src/AnimationState.cs
+++ b/spine-csharp/src/AnimationState.cs
@@ -1214,6 +1214,12 @@ namespace Spine {
 			get { return nextTrackLast != -1; }
 		}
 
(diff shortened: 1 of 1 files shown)


NathanSweet:
At first glance this doesn't appear makes sense. `next` makes up a linked list of queued animations. The `next` animation is _never_ applied until it's made the current animation and is no longer "next". The only animations that are applied are 1) the current animation for a track, and 2) the animations being mixed out.

TrackEntry `wasApplied` is true if the entry has ever been applied. It does not always return true when there is `next` animation, unless there is a bug. `nextTrackLast` starts as -1 and is only set when the TrackEntry is applied, so there doesn't appear to be a bug.

To avoid an [xy problem](https://xyproblem.info/) it may be better to start from your problematic use case.

TyelorD:
> At first glance this doesn't appear makes sense. `next` makes up a linked list of queued animations. The `next` animation is _never_ applied until it's made the current animation and is no longer "next". The only animations that are applied are 1) the current animation for a track, and 2) the animations being mixed out.
> 
> TrackEntry `wasApplied` is true if the entry has ever been applied. It does not always return true when there is `next` animation, unless there is a bug. `nextTrackLast` starts as -1 and is only set when the TrackEntry is applied, so there doesn't appear to be a bug.
> 
> To avoid an [xy problem](https://xyproblem.info/) it may be better to start from your problematic use case.

Hey Nathan, my use-case is outlined [here](https://esotericsoftware.com/forum/d/24647-animation-doesnt-change-on-frame-setanimation-is-called). A better name for this function should actually be "WillNextBeAppliedOnUpdate" since it tells us if the next animation in the queued animations is going to be applied on the next update call.

NathanSweet:
That is a very long forum thread, the use case is not easily extracted.

The new method name makes more sense, if a bit wordy. I'm not a big fan of methods starting with `will`. Maybe `isNextReady`. Any better ideas?

I think the logic should match the check in `update` exactly:
```
return next != null && nextTrackLast - next.delay >= 0;
```
Can you please test this to ensure it still works for your use case?

Another idea is `isReady`, which would return true if this TrackEntry has a previous entry whose track time has advanced enough that this entry will become current on the next update:
```
return previous != null && previous.nextTrackLast - delay >= 0;
```
This would most commonly be used like `entry.next.isReady()`. The method name is nice, though with the downside that the caller may want to null check `next`.

TyelorD:
> That is a very long forum thread, the use case is not easily extracted.

That's a good point, plus it was quite a while ago that we had discussed it together. I'll quote the relevant details here:

> A high level overview of what we are doing is:
> - Character starts in their Idle state, and the skeleton is posed to their "idle" animation.
> - A timer fires after X seconds of Idling, setting their state to Punching
> - Upon entering the Punching state we use SetAnimation("punching")
> - After the "punch" track's completion callback we set the Character to their PunchReset state.
> - Upon entering the PunchReset state we use SetAnimation("punch_reset")
> - After the "punch_reset" track's completion callback we set the Character back to their Idle state.
> - Upon entering the Idle state we use SetAnimation("idle") and reset the timer to fire again after X seconds.
> 
> Here the "punching" animation has an event called "vulnerable_on" which we listen for to set a flag to the game that the character themselves can be punched during this portion of the animation. There is another event in "punch_reset" called "vulnerable_off" to reset this flag.
> 
> However we also have a system where when the AnimationState calls its Start callback, we automatically reset the vulnerability flag to "off", this allows us for most animations to not add a "vulnerable_off" event explicitly at the end of many animations.
> 
> But in the case above we actually want the vulnerability flag to persist between the "punching" and "punch_reset" animations until "punch_reset" turns it off with the "vulnerability_off" event, so at frame 0 of "punch_reset" we actually call another "vulnerable_on" event to counter the automatic reset. Herein is where the problem lies though, since as you pointed out the AnimationState is only applied once per frame there is a single frame delay between when the AnimationState calls its Start callback and when the "punch_reset" track is actually applied to the AnimationState. This in turn causes a single frame where the vulnerability flag is set to "off" between the automatic reset when the AnimationState calls its Start callback and when the "punch_reset" is applied to the AnimationState (and thus fires its frame 0 events) on the following Update.


> The new method name makes more sense, if a bit wordy. I'm not a big fan of methods starting with `will`. Maybe `isNextReady`. Any better ideas?

That's good feedback, and I felt the same way. I like `isNextReady` it is clear (at least to me) and very concise! The only other ideas I have off the top of my head are more wordy and verbose than your idea.

> I think the logic should match the check in `update` exactly:
> 
> ```
> return next != null && nextTrackLast - next.delay >= 0;
> ```
> 
> Can you please test this to ensure it still works for your use case?

After a few quick smoke tests it does seem like that logic still works the same for our use case. I've gone ahead and updated the PR accordingly.

Thanks Nathan!

PR #2545, opened 2024-06-06
Update Attachment.ts Fix Nan

verketh:


NathanSweet:
Why?

verketh:
[spine.zip](https://github.com/user-attachments/files/15711351/spine.zip)
This is the spine where the bug occurs. The switching animation can be reproduced. The hat part will cause rendering errors due to nan

badlogic:
I can not repro this in 4.2. Closing this out.

PR #2578, opened 2024-07-14
Update SpineAnimationTrack.cpp Fix: track_index step cannot be 0.

indubitablement2:
Fix: `Invalid PROPERTY_HINT_RANGE with hint "0,256,0": Step cannot be 0.`


badlogic:
Cheers!

PR #2599, opened 2024-08-13
replaces setTransform on slotObject

GordonTombola:
Sorry only just noticed this also:

setTransform on pixi displayObjects performs:
```
this.scale.x = !scaleX ? 1 : scaleX;
this.scale.y = !scaleY ? 1 : scaleY;
```
meaning any transformations to scale 0 will become scale 1


davidetan:
Thanks for noticing this!
Slot objects are a recent addition, so I'm not surprised there are small issues with them.

PR #2626, opened 2024-09-09
Fix `SpinePlayerConfig` typing and `dispose()` error before init

RaenonX:
Basically copy pasted what 4.2 has with small modification for `SpinePlayerConfig`.

Side note though, I think the typing for 4.2 on `preserveDrawingBuffer` for `SpinePlayerConfig` is also incorrect. The comment marks it as optional (and from the usage it seems optional) but 4.2 has it as required.


davidetan:
Related to this forum thread: https://esotericsoftware.com/forum/d/26766-unable-to-load-spine-fils-using-spine-web-player/18

I'll merge this and make a release later today.

PR #2644, opened 2024-09-26
Fix possible NAN in IkConstraint.cpp

archilanfedorov:
additional negative sqrt protection, float expressions may be not exact on android NDK 25.2.9519653 devices Oppo Find x3, Huawei lite M5

image


badlogic:
Thanks!

PR #2646, opened 2024-09-29
Make String::startsWith and String::lastIndexOf const.

chemodansama:
The `String::startsWith` and `String::lastIndexOf` methods were not marked `const` which makes it impossible to use them when we only have constant String object.


badlogic:
Thanks!

PR #2648, opened 2024-10-02
[spine-unity] SkeletonRenderTexture set dynamic GameObject layer same as original Renderer

mavlac:
When using the example `SkeletonRenderTexture` Component, new GameObject is created in hierarchy to hold Quad with RenderTexture Renderer. Sorting layer and sorting order values are mimicking the original Renderer as expected, but the GameObject Layer is not and kept Default.

Please consider this tiny touch, so the dynamically created Renderer is as close to original Renderer and so is for example being visible to the same Cameras when using layer mask culling.


HaraldCsaszar:
Thanks for reporting and sorry for the late reply, I've overlooked this pull request completely!
A new unitypackage has been published including this change:

mavlac:
No trouble at all, many thanks - still going to be very appreciated.

PR #2649, opened 2024-10-03
[csharp] Update csproj to Framework 4.8 and fix missing files

brogan89:
- Updated the csproj to a supported version of .NET Framework as v4.0 is no longer supported and I didn't want to download it 😁
- Added missing files so the project could compile.

Hopefully that's all good. It seems like the core C# stuff is very coupled to Unity/MonoGame. I'm kind of hoping I can leverage this to use in Godot-Mono so I don't need to use custom compiled Godot Engine.


badlogic:
spine-csharp isn't really coupled to Unity/Mono. That said, you'll have to basically rewrite what spine-godot does, including things like rendering of meshes, asset importing, and so on.

I'm finishing up the spine-godot GDExtension work as we speak, which will allow you to use spine-godot in any Godot engine build. That said, GDExtension and C# aren't best buddies yet. There's no way to generate C# bindings for a GDExtension, so you would have to use the "reflection" like ClassDB and Object APIs.

@HaraldCsaszar can you sign off on this?

HaraldCsaszar:
@brogan89 Sorry that I've overlooked this earlier. The spine-csharp.csproj changes slipped through as spine-monogame.csproj explicitly references the core-runtime files, thanks for the pull request. Increasing the framework version to 4.8 does indeed make a lot of sense as well.

Regarding spine-csharp being coupled to spine-unity or spine-monogame: this is not the case. Did you get that impression due to the additional files like the `.meta` files and the `package.json` file used by Unity, or did you see any concrete problem? Note that the `.meta` and `package.json` files are just allowing easier integration and are ignored by other game toolkits or runtimes. There is no specific Unity or monogame related code in the spine-csharp core runtime.

brogan89:
> @brogan89 Sorry that I've overlooked this earlier. The spine-csharp.csproj changes slipped through as spine-monogame.csproj explicitly references the core-runtime files, thanks for the pull request. Increasing the framework version to 4.8 does indeed make a lot of sense as well.
> 
> Regarding spine-csharp being coupled to spine-unity or spine-monogame: this is not the case. Did you get that impression due to the additional files like the `.meta` files and the `package.json` file used by Unity, or did you see any concrete problem? Note that the `.meta` and `package.json` files are just allowing easier integration and are ignored by other game toolkits or runtimes. There is no specific Unity or monogame related code in the spine-csharp core runtime.

Yeah it was just the `.meta` files mostly and the `package.json`. There is also the `#define IS_UNITY`.
But yeah other than that it works perfectly fine as a standalone. In fact, I even bumped the version to dotnet8 locally and it works fine, I got it rendering with OpenGL using Silk.NET. I thought about making a PR for that as a demo of how to use Spine C# outside of Unity/MonoGame. If you're interested? (off topic sorry)

Regarding Godot-csharp GDExentions sounds like best way to go. Its working well for GodotSteam with the [C# bindings](https://github.com/LauraWebdev/GodotSteam_CSharpBindings), it uses `ClassDB` to wrap the C++ plugin as well, I think that's a perfectly fine way to do it.

PR #2653, opened 2024-10-09
[ts] Add sources to source maps via TS configuration option

### LINKED ISSUE #2652: [ts] Missing `"sourcesContent"` field in source maps
Source maps are generated for .ts files in `spine-ts` folder with a `"inlineSourceMap": true,` setting (see `spine-ts/tsconfig.base.json`). Since separate source map file are not meant to be generated, I suppose that the inline source map should contain a `"sourcesContent"` field with contents of the source .ts files, so that source maps work correctly.

This could be achieved by adding ["inlineSources": true](https://www.typescriptlang.org/tsconfig/#inlineSources) to tsconfig.base.json.

alvov-evo:
Fixes #2652

### DIFF
--- a/spine-ts/tsconfig.base.json
+++ b/spine-ts/tsconfig.base.json
@@ -4,6 +4,7 @@
     "module": "ESNext",
     "noImplicitAny": true,
     "preserveConstEnums": true,
+    "inlineSources": true,
     "inlineSourceMap": true,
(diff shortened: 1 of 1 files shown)


davidetan:
We never noticed this since we usually develop locally where source files are present.
Thanks!

PR #2662, opened 2024-10-25
[Spine-Android]FIX: clipEnd is not called  when attachment is null.

mrrobot97:
If the attachment of the slot which has **End Slot** properties set is **null**, then _clipEnd_ will not called ever, which will cause rendering issue.


badlogic:
Good catch, thank you for the pull request! I'll release a new version today.

PR #2669, opened 2024-11-06
[ios] Fix the issue where the SkeletonBounds instance couldn’t be initialized

kikiloveswift:
`SkeletonBounds` has internal access, which is determined by the Python code. Because of that, we need to add a static method in `Spine.Generated+Extensions.swift` to create an instance.


badlogic:
Thanks!

Issue #2675, opened 2024-11-08
[ts][pixi-v8] Slot objects render with a delay


SerG-Y:
Hi guys, first of all, thanks for adding Pixi v8 support, great job!

I encountered a few issues when migrating from PixiJS v7 to v8 (all of these worked fine with PixiJS v7 and pixi-spine v7):
- Slot objects render with a delay. You can notice the "PLAY" text blinking, as the slot object becomes visible 1–2 frames later than the animation itself.
- Unable to destroy the animation within the complete callback.
- Unable to create a new animation within the complete callback.

I have prepared an example demonstrating all these issues:
v8:
https://codesandbox.io/p/sandbox/9yh9cf

working example v7: 
https://codesandbox.io/p/sandbox/k8jdnc

davidetan:
Hello and thanks for trying the new runtime!

The issues seem quite clear and I have already an idea about what could cause the the lasts two.

I'll let you know as soon as I fixed them!

davidetan:
We just released 4.2.65 that should address all three issues.
You probably need to move the destroy after the new button has been created to avoid an empty frame.

Serhiy-Y:
@davidetan Thanks a lot!
Also, I saw your commit [fb3855a](https://github.com/EsotericSoftware/spine-runtimes/commit/fb3855ad6588d5d8ed5cfb5991a13fd858af9a29), great job!

PR #2679, opened 2024-11-12
[ts][pixi-v8] Moved state.apply and UWT right after time updates.

davidetan:
This commit should address the problems mentioned into #2675.

Here we moved the state apply, the skeleton updateWorldTransform and the slot objects update together with the state and skeleton update.
The whole state is now always in sync.
Methods where renamed like this:

- `_updateState` -> `_updateAndApplyState`
- `_applyState` -> `_validateAndTransformAttachments`

However, this has the drawback that the constructor cannot invoke the `_updateState`, now `_updateAndApplyState`, otherwise the setup pose is always rendered.
That call was probably there to calculate bounds. It was removed.

Instead, the very first time `updateBounds` is called - for example when the user invoke `getBounds`, the `_updateAndApplyState` is called to guarantee the bounds are calculated with a state that reflects what the user wants.

@GoodBoyDigital What do you think about this modifications?

### DIFF
--- a/spine-ts/spine-pixi-v8/example/control-bones-example.html
+++ b/spine-ts/spine-pixi-v8/example/control-bones-example.html
@@ -53,12 +53,12 @@
         // from one animation to the next.
         stretchyman.state.data.defaultMix = 0.2;
 
+        // Set animation "run" on track 0, looped.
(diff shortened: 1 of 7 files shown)


GoodBoyDigital (approved):
nicely done @davidetan 👍

PR #2693, opened 2024-11-22
[ts][pixi-v7] Fix incorrect pixiObject scale in updatePixiObject

Hafaux:
This commit resolves a bug where both the X and Y components of the pixiObject's scale were incorrectly set to the bone's worldScaleX, instead of applying the correct scaling for each axis.

Additionally, I have a question (apologies if this isn't the appropriate place): Is it intentional that a pixi object added using addSlotObject does not apply the slot's transformations and only the parent bone's transform? The method name suggests it would, so I'm wondering if I'm misunderstanding something. Thanks for the quick updates :)


davidetan:
Good catch!

General questions can be asked on [our forum](https://esotericsoftware.com/forum), but it's completely ok to ask something short and related into the PR.
A slot does not have a transform. Slots in Spine are conceptual and are used to organize and group attachments, as well as to define the draw order.

PR #2694, opened 2024-11-22
[cpp] avoid null pointer deference in PhysicsConstraintTimeline

bendmorris:
Compiling spine-cpp with mingw and -Werror results in:

```
/home/ben/Dev/march/krit/spine-runtimes/spine-cpp/spine-cpp/src/spine/PhysicsConstraintTimeline.cpp:98:75: error: ‘this’ pointer is null [-Werror=nonnull]
   98 |                                 if (constraint->_active) constraint->reset();
      | 
```

This looks like a real issue as `constraint` is definitely null here. The intent seems pretty clear though so here's a quick fix - feel free to throw it away and fix it a different way.

### DIFF
--- a/spine-cpp/spine-cpp/src/spine/PhysicsConstraintTimeline.cpp
+++ b/spine-cpp/spine-cpp/src/spine/PhysicsConstraintTimeline.cpp
@@ -95,6 +95,7 @@ void PhysicsConstraintResetTimeline::apply(Skeleton &skeleton, float lastTime, f
 		else {
 			Vector<PhysicsConstraint *> &physicsConstraints = skeleton.getPhysicsConstraints();
 			for (size_t i = 0; i < physicsConstraints.size(); i++) {
+                                constraint = physicsConstraints[i];
 				if (constraint->_active) constraint->reset();
 			}
 		}


spine-cpp/spine-cpp/src/spine/PhysicsConstraintTimeline.cpp
```diff
@@ -95,6 +95,7 @@ void PhysicsConstraintResetTimeline::apply(Skeleton &skeleton, float lastTime, f
 		else {
 			Vector<PhysicsConstraint *> &physicsConstraints = skeleton.getPhysicsConstraints();
 			for (size_t i = 0; i < physicsConstraints.size(); i++) {
+                                constraint = physicsConstraints[i];
 				if (constraint->_active) constraint->reset();
```
hsdk123, line 99: I'd think you also want to `if (constraint && constraint->_active)`

bendmorris: There isn't a null check above either, or in other places where these are used. My assumption would be that `getPhysicsConstraints` returns a vector of non-null pointers so if you're within its size the check isn't necessary. If that's not the case happy to add the check.

badlogic:
Excellent catch! Sorry I overlooked your PR. I've now manually fixed this up. Thanks for reporting!

PR #2695, opened 2024-11-26
[TS] Update the world transform of each bone in `updateWorldTransformWith`

smaren:
The LibGDX version of `updateWorldTransformWith` aldready does this, so I added it here as well.


davidetan:
Good catch, thank you!
I had to add an additional line of code to stick with the reference runtime, so it was easier to just do a commit with everything.

PR #2698, opened 2024-11-28
[cmake] Add SPINE_SET_COMPILER_FLAGS

hsdk123:
The spine library setting compiler flags may not be wanted by the user using this library.
Adding an option to make it adjustable.


badlogic:
Thanks!

hsdk123:
@badlogic 
https://github.com/EsotericSoftware/spine-runtimes/pull/2716
Fixed an error - if you could look at this as well, would be great!

PR #2717, opened 2024-12-18
Fix: pixi v8 bug in validate renderables

GoodBoyDigital:
I notived that my inital implementation was should have been reseting `spineAttachmentsDirty` and
`spineTexturesDirty` each render.

This caused two issues:
- once `spineTexturesDirty` is forever true so it will always cause a rebuild of the scene.
- it can be the case that `spineAttachmentsDirty` is set to true and then set to false in the `_validateAndTransformAttachments` function. This will cause an issue if slots have changed, but not been rendered yet.

Easy enough fixes.. once I figured out where to look!


davidetan:
Cool, thanks for this! Released in `4.2.70` 🚀

PR #2720, opened 2024-12-27
[ios] Using SIMD to replace vector_float types can boost CPU performance on mid-range models.

kikiloveswift:
I've been working on further optimizing Spine performance lately. For Spine files with a lot of vertices on mid-range devices with moderate performance, using SIMD significantly optimizes CPU usage, reducing it by about 10 to 20%. However, for files with a typical vertex count on high-end flagship devices, the optimization gains from SIMD aren't as pronounced, only reducing CPU usage by about 4%.


badlogic:
This is excellent, thank you!

I'm actually surprised, that the compiler doesn't auto-vectorize most of this, and that changes to the shader have an impact on CPU usage.

PR #2736, opened 2025-01-21
[cpp] Fix mixShearY assignment in SkeletonJson::readAnimation

akhuang:
mixShearY2 was not assigned to mixShearY. Is this a bug?


badlogic:
Thanks! Two excellent catches. I've fixed this manually and pushed the fix to the 4.2 branch. I have not added the SP_ATLAS_UNKNOWN_FORMAT constant. That's also defined in spine-c, where it actually isn't used. The Spine Runtimes aren't super defensive in terms of parsing, and mostly assume that the data to be read is valid data.

PR #2755, opened 2025-02-18
Bulletproof Skin.attachAll()

ljurow:
attachments Array can have null entries.


davidetan:
This was defeinitely needed. Good catch!

davidetan:
Released in 4.2.4

Issue #2753, opened 2025-02-14
[Godot] Crash when `Visible Collision Shapes` is enabled


misaki-eymard:
See the following thread for the reproduction steps and the Spine and Godot projects for reproduction:
https://esotericsoftware.com/forum/d/28136-godot-game-crash-if-collission-debug-is-enabled-with-spinesprite

PR #2761, opened 2025-02-21
fix godot crash when Visible Collision Shapes is enabled

ferdyferdy9:
Fixes the issue of #2753

The crash is caused by the game being unable to obtain EditorInterface singleton during runtime


badlogic:
Oh, wonderful! Thank you very much!

PR #2772, opened 2025-03-02
[iOS] update Swift inteface for more fine grained control(like external atlas loading)

pbk20191:
- move instance method to static method which does not require spine cpp pointer
- implement equality and hashing based on cpp pointer so that it can be stored in collection and compare it
- expose Atlas image count property so that the user can load the atlas resource lazily

- User can now Create Atlas by using Altas static function, and fetch whole resource path by iterating count of atlas page count

```swift
let atlas = Atlas.load(atlasData:"atlas txt file content string!")
if let error = atlas.error {
    throw SpineError(error)
}
let count = Int(atlas.imagePathCount)
let atlasImagePaths:[String] = (0..<count).map{ atlas.getImagePath(index: Int32($0))! }

// load the desired png only when needed maybe?
```


badlogic:
Thanks!

PR #2774, opened 2025-03-04
[godot] Allow double precision Godot Builds

granitrocky:
This swaps out the `<float>` stuff for `<real_t>`. It also updates the ResourceLoaders to work with godot 4.4.

I haven not tested this with a single precision build, which is definitely on my radar. There's probably some cleanup necessary in here, too.

### DIFF
--- a/spine-godot/spine_godot/SpineBone.cpp
+++ b/spine-godot/spine_godot/SpineBone.cpp
@@ -140,27 +140,27 @@ Vector2 SpineBone::parent_to_world(Vector2 local_position) {
 	return Vector2(x, y);
 }
 
-float SpineBone::world_to_local_rotation(float world_rotation) {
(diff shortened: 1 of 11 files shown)


badlogic:
I'm not sure this will work as expected. The underlying spine-cpp runtime that is called in e.g. the setters of SpineBone, still use float. If you push in say a skeleton position with a double value that's bigger than what a float can handle, funny things will happen.

Changing spine-cpp to use double precision is not something we can do in the short term, as it's quite a lot of work.

granitrocky:
Thanks for taking a look @badlogic

It's been a little while since I took a look at this, but iirc I actually did cast everything to float when interacting with the underlying spine library.

But I understand that maintaining this would probably not be something you're interested in. I'll keep it locally for now and if anyone is interested they can comment on my fork.

Issue #2777, opened 2025-03-05
[iOS] Issue with Cxx module on Xcode 16.2


janpodmolik:
Hi, I recently upgraded to Xcode 16.2, and I started experiencing the following error when building for iOS:

```
<unknown>:0: error: compiling for iOS 13.0, but module 'Cxx' has a minimum deployment target of iOS 16.0: /Applications/Xcode-16.2.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/iphonesimulator/Cxx.swiftmodule/arm64-apple-ios-simulator.swiftmodule
```

I have already set my minimum deployment target to iOS 18, so the error message about iOS 13 seems misleading. This issue did not occur before updating to Xcode 16.2.

I am wonder if anyone else has encountered this problem.
Is this a known issue? Is there any planned fix or workaround?

Thanks in advance for any help!

pbk20191:
This is duplicated issue of #2712

badlogic:
This is fixed via #2780, thanks @pbk20191 !

Issue #2712, opened 2024-12-14
[iOS] Cxx and CxxStdlib built for iOS 16.0 and above only


yihui-hu:
With the release of Xcode 16.2 on Dec 11, the minimum iOS deployment target for Swift/C++ mixed projects has been increased to 16.0 (see [this swift forum post](https://forums.swift.org/t/xcode-16-2-beta-seems-increase-the-min-deploy-to-16-0-for-swift-cpp-mixed-project/76221) for more detail).

Since the `Package.swift` in the repo targets iOS 13, my project has been unable to build.

The workaround that I'm using is downgrading to an older Xcode binary until this is resolved.

nosferatu500:
Seems like this is Xcode 16.2 issue (simulator only)

[Known Issues](https://developer.apple.com/documentation/xcode-release-notes/xcode-16_2-release-notes#Swift)
```
Xcode targets using Swift/C++ interoperability and having a deployment target of less than iOS 16 or tvOS 16 may not build for iOS or tvOS simulators. (141232269)

[No Workaround]
```

Should be fixed [here](https://github.com/swiftlang/swift/pull/78242)

PS: No any issues on a real device (on my end).

badlogic:
We still need to support older versions of iOS, as such the target in our Package.swift needs to stay at 13 I'm afraid. Let's hope the fix in Swift gets released soonish.

badlogic:
Fixed via #2780

PR #2780, opened 2025-03-07
[iOS] fix compile failure using SPM with recent xcode

pbk20191:
since SpineCppLite expose c-interface only and does not expose cpp interface

c++ interperability is no-op and just causing compiler crash.
So we could add linker flag to ensure SpineCppLite is build with C++ while removing c++ interperability mode

I have tested it using Spine example project, and it will solve #2777 , #2712


badlogic:
Excellent fix! Thanks.

PR #2781, opened 2025-03-09
Fix memory leak of Vector::setSize

dumganhar:


badlogic:
Cheers!

dumganhar:
@badlogic Thanks for the merge. Would you cherry-pick this patch to other branches?

PR #2785, opened 2025-03-13
[csharp] Fix bug in SkeletonBinary.SkeletonInput.ReadInt

ww-rm:
The `ReadInt` method in C# should align with the behavior of the C version. Here, an unsigned right shift should be used to ensure that the higher bits are consistently set to 0.

Expected:

https://github.com/EsotericSoftware/spine-runtimes/blob/18d30af5a9cd2429a16e003b85dc844a8f6f2ea9/spine-c/spine-c/src/spine/SkeletonBinary.c#L152

### DIFF
--- a/spine-csharp/src/SkeletonBinary.cs
+++ b/spine-csharp/src/SkeletonBinary.cs
@@ -1287,7 +1287,7 @@ namespace Spine {
 						}
 					}
 				}
-				return optimizePositive ? result : ((result >> 1) ^ -(result & 1));
+				return optimizePositive ? result : ((int)((uint)result >> 1) ^ -(result & 1));
 			}
 
 			public string ReadString () {


HaraldCsaszar:
@ww-rm Thanks for reporting, good catch!

A new spine-unity 4.2 unitypackage is available for download:

ww-rm:
@HaraldCsaszar Thanks for the response!

This bug seems to exist across all historical versions of the C# implementation. Hope it hasn't caused any severe problem.

HaraldCsaszar:
> This bug seems to exist across all historical versions of the C# implementation. Hope it hasn't caused any severe problem.

@ww-rm Thanks, I've noticed that as well. It's highly unlikely, as the codebranch is only used with user event when setting the event integer value. This value would have been noticed rather easily when it's off.

PR #2787, opened 2025-03-17
[haxe] When SkeletonSprite is disposed, clean up listeners

ljurow:
If you don't cleanup listeners, then callbacks get call for an object that is already thrown away.


davidetan:
Thanks for noticing it! I'll fix it and release a new package version tomorrow :)

davidetan:
I had a couple more things to do, so I made a separate commit: 34ccf846a9cfec71dd17b9bf3ef733f240460900.

Available in 4.2.6. Thanks again!

PR #2789, opened 2025-03-20
SkeletonAnimation running at multi-threads mode

suntabu:
This is just an early version of multi-threads mode craft.

I am sharing this for anybody who needs parallelization spine runtime.

Any suggestions and optimizations are welcome.

At last, due to my limited programming skills, this solution may have many issues. Please use it with caution, as it works fine for our project team's needs.

### DIFF
--- a/spine-unity/Assets/Spine/Runtime/spine-unity/Mesh Generation/MeshGenerator.cs
+++ b/spine-unity/Assets/Spine/Runtime/spine-unity/Mesh Generation/MeshGenerator.cs
@@ -819,7 +819,9 @@ namespace Spine.Unity {
 			int totalVertexCount = instruction.rawVertexCount;
 
 #if LINEAR_COLOR_SPACE_FIX_ADDITIVE_ALPHA
-			bool linearColorSpace = QualitySettings.activeColorSpace == ColorSpace.Linear;
+			// bool linearColorSpace = QualitySettings.activeColorSpace == ColorSpace.Linear;
+			//TODO::: we only use linear color space in our projects
+			bool linearColorSpace = true;
 #endif
 			// Add data to vertex buffers
 			{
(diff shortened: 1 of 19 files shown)


HaraldCsaszar (commented):
You should also be setting `isPlaying = Application.isPlaying;` in Awake, otherwise utility components might be indirectly accessing `IsPlaying` during initialization and see it as false.

In general you can wrap these in #if UNITY_EDITOR blocks and exclude them in the built executable, as it's only relevant in the editor.

spine-unity/Assets/Spine/Runtime/spine-unity/Mesh Generation/MeshGenerator.cs
```diff
@@ -819,7 +819,9 @@ public void BuildMeshWithArrays (SkeletonRendererInstruction instruction, bool u
 			int totalVertexCount = instruction.rawVertexCount;
 
 #if LINEAR_COLOR_SPACE_FIX_ADDITIVE_ALPHA
-			bool linearColorSpace = QualitySettings.activeColorSpace == ColorSpace.Linear;
+			// bool linearColorSpace = QualitySettings.activeColorSpace == ColorSpace.Linear;
+			//TODO::: we only use linear color space in our projects
+			bool linearColorSpace = true;
```
HaraldCsaszar, line 824: Obviously, whoever uses this code will want to adjust these lines accordingly.

HaraldCsaszar:
Thanks for sharing your code, others will be happy to include it in their projects until the official solution is released. I've quickly skipped over your code and it does not look bad overall. While it's very limited, I'm sure it will still benefit some users. I've commented on an obvious potential issue with `isPlaying`. Obviously also setting `bool linearColorSpace = true;` is something other users shall adjust accordingly to their project needs.

Closing this pull request now.

PR #2799, opened 2025-04-01
Remove SkeletonSprite from the Juggler on dispose

ljurow:
When we dispose of the SkeletonSprite, we should remove it from the juggler - since this is the default thing that would be calling advanceTime().  If the SkeletonSprite is not in the juggler, nothing happens, but if we don't remove it - we now have a memory leak (and we are calling advanceTime on a item that it disposed).

You could make an argument then when we create a SkeletonSprite, it should be added to the Starling juggler automatically.  However, I can see arguments to NOT do this (i.e. you aren't using the default juggler, or you are using something else to call advanceTime).

### DIFF
--- a/spine-haxe/spine-haxe/spine/starling/SkeletonSprite.hx
+++ b/spine-haxe/spine-haxe/spine/starling/SkeletonSprite.hx
@@ -423,6 +423,7 @@ class SkeletonSprite extends DisplayObject implements IAnimatable {
 			_state = null;
 		}
 		if (_skeleton != null) _skeleton = null;
+		if (null != Starling.current && null != Starling.current.juggler) Starling.current.juggler.remove(this);
 		super.dispose();
 	}
 }


davidetan:
Hello, and thanks for the PR!  

I appreciate the concern about potential memory leaks, but I don’t think this is actually a memory leak issue.  
If the `dispose` method is called, the `AnimationState state` is set to `null`.  
Then, on the next `advanceTime` invocation, `_state.update(time)` will be executed, which will inevitably crash the application.  

That said, as you pointed out, the default juggler is not the only way to advance the `SkeletonSprite`'s time.  
Other cases include:  

1. **User-instanced jugglers** – The user is responsible for manually removing the `SkeletonSprite`.  
2. **Manual `advanceTime` calls** – The user should simply stop calling `advanceTime`.  
3. **Custom structures for advancing time** – Same as the first case; the user must handle it manually.  

In other spine runtimes, when a game object doesn’t have an implicit update method called every tick (like in the Haxe Flixel runtime), we automatically add it to the default juggler—unless the user explicitly sets an auto-update property to `false` (as in the Pixi runtimes).  
For consistency, we should do the same here so that most users don’t need to worry about adding a Spine game object to the default juggler.  
It makes no sense to use a Spine game object without animating it, and if a user doesn’t want automatic updates, they can simply set auto-update to `false`.  

> You could argue that `SkeletonSprite` should be added to the Starling juggler automatically upon creation.  
> However, I also see reasons **not** to do this—such as users relying on a different juggler or manually calling `advanceTime`.  

Well, I just made that argument myself! :joy:  
I’ll probably consider making this change in the 4.3-beta branch, as it would be a breaking change.  

Beyond all of this, there’s another issue with game object disposal right now. If the object is not removed from the stage, another crash will occur.  

Ideally, `dispose` should also:  
- Remove the game object from the stage.  
- Remove the game object from the juggler.  

Removing it from the stage is simple, we just call `removeFromParent()`.  
For the juggler removal, after digging into the Starling code (which has very little documentation or discussion online), I found that a `DisplayObject` can dispatch a `REMOVE_FROM_JUGGLER` event, which automatically removes it from any juggler it was added to.  
This is super convenient, as it resolves both the default and user-instanced juggler cases.  

For manual `advanceTime` calls, nothing needs to be done, and for custom time-advancing structures, it’s clearly the user’s responsibility.  

I implemented this in commit `4ea6ee9` and will release it in Haxe 4.2.7.

davidetan:
4.2.7 released :)

PR #2800, opened 2025-04-05
[cpp] fix bound computation ignores clipping

pbk20191:
cpp bound calculation ignores Clipping while other platform does not

https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/spine-libgdx/spine-libgdx/src/com/esotericsoftware/spine/Skeleton.java#L728
```java
	} else if (attachment instanceof ClippingAttachment && clipper != null) {
		ClippingAttachment clip = (ClippingAttachment)attachment;
		clipper.clipStart(slot, clip);
		continue;
	}
```


badlogic:
Cheers!

PR #2806, opened 2025-04-08
expose Spine Package to all apple platform while preserving current behavior

pbk20191:
This PR exposes Spine Package to all available current apple platform
- only iOS platform can import original `Spine` module
- other platform can import `SpineCppLite` and `SpineShadersStructs` but not `SpineModule`
- In this way user can manage multiplatform project while using Spine-runtime

- This pr replace library product named `Spine` to `SpineModule`  from `Spine`  target, and `SpineModule` reexport all the modules.

So the existing swift package library user has no negative impact.

This PR is tested using "Spine example Project"

`SpineCppLite`, `SpineShaderStructs` was compiled successfully in xcode 16.2 with all available platform
(macOS-universal, macCatalyst-universal, tvOS, tvOS-simulator, watchOS, watchOS simulator, visionOS, visionOS-simulator)


badlogic:
Awesome, cheers!

PR #2808, opened 2025-04-09
[iOS] merge shader header and disable cpp-interp in example project

pbk20191:
This merge ShaderTypes reusing same header file


badlogic:
Cheers!

PR #2816, opened 2025-04-17
[iOS] Spine fix Metal deadlock

pbk20191:
Current implementation may fall into deadlock

- CAMetalLayer is configured with drawable timeout, and CAMetalLayer is run out of Drawable 
- CAMetalLayer is added to the window with frame size of zero or incorrect layout constraint (currentRenderPassDescriptor is null)


badlogic:
Thank you!

PR #2826, opened 2025-04-26
[ts][webgl] SpineCanvas constructor config.webglConfig condition correted

Zormolo:
grafik

Condition of the config.webglConfig ignored the parameter. After condtion is fixed the paramter is set properly.


davidetan:
We cherry picked the previous commit. I've released a new version with the fix (4.1.55).

PR #2829, opened 2025-04-27
[ts] use skeletonData as a global viewport alternative

chejen:
[This doc](https://esotericsoftware.com/spine-player) explains that there's an automatic viewport feature by default. For instance, in [example.html](https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/spine-ts/spine-player/example/example.html), without setting a global viewport, the spineboy and raptor characters will shrink when switching animations from 'walk' to 'jump'.

[original.webm](https://github.com/user-attachments/assets/079c9962-6f42-485e-9f54-09a5c4e9aab1)

While setting a global viewport resolves this, it's a manual step. Ideally, we could leverage the loaded skeletonData to avoid viewport updates whenever the Skeleton file is modified. So I'm wondering if the following could be an alternative to manual setting x, y, height and width:
```js
var jsControlledPlayer = new spine.SpinePlayer("container-raptor", {
	skeleton: "assets/raptor-pro.json",
	atlas: "assets/raptor-pma.atlas",
	animation: "walk",
	viewport: {
		useSkeletonData: true,
	},
});
```

[useSkeletonData.webm](https://github.com/user-attachments/assets/e8b36f56-29d1-42b7-a69c-734d3cafa9e2)

The characters don't shrink even if we don't explicitly set x, y, height and width ourselves.

### DIFF
--- a/spine-ts/spine-player/src/Player.ts
+++ b/spine-ts/spine-player/src/Player.ts
@@ -120,6 +120,9 @@ export interface SpinePlayerConfig {
 
 		/* Optional: Viewports for specific animations. Default: none */
 		animations?: StringMap<Viewport>
+
+		/* Optional: Whether to use the x, y, height and width in skeletonData as global viewport. Default: false */
+		useSkeletonData?: boolean
 	}
 
 	/* Optional: Whether the canvas is transparent, allowing the web page behind the canvas to show through when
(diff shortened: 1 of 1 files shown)


davidetan:
The skeleton's x/y/width/height define the position and size of the AABB for the skeleton’s attachments as they appear in the setup pose in the editor.

This rectangle might make sense for a non-animated preview or something similar, but it generally doesn't fit animated poses well.  
For example, in the raptor example, the `walk` animation doesn't even fit within the viewport, as you can see in this screenshot (spineboy's hair or the raptor's mouth extend outside):  
image

Moreover, even though it's a manual step, copying and pasting those values is a trivial operation.

For these two reasons, I don't think it's worth adding a new config parameter.

It might make more sense to add a parameter to create a fixed viewport that fits a list of animations, similar to what we are doing in our experimental Spine web component ([see here](https://github.com/EsotericSoftware/spine-runtimes/blob/ts-webgl-overlay/spine-ts/spine-webgl/src/SpineWebComponentWidget.ts#L483)).
Once we merge the Spine web component, I'll consider adding this option to the player as well.

In any case, thanks for your idea and PR!

chejen:
@davidetan Thank you for the reply.

I agree with you on that point. That said, there are a few reasons why I'm suggesting this approach — or something along these lines.

1. We've used the same animation list across different platforms, but the web platform is the only one where we have to manually set the position and size for the animations. On iOS and Android, this happens automatically.

2. While copying and pasting the data is straightforward, there's a risk that when designers update the animations, the size and position of the viewport could change — and when that happens, it's usually only the web developers who need to update the code and redeploy to handle those changes.

Do you have any suggestions before we move forward with rolling out the experimental Spine web component? Thank you!

PR #2830, opened 2025-04-28
Fix the inconsistent behavior between Spine vector's setSize and std::vector's resize, which causes the Spine vector to allocate additional space upfront.

bofeng-song:
企业微信截图_17458243154941

As shown in the screenshot, when the vector is empty, the first call to resize results in capacity and size being equal, meaning no additional capacity is allocated.


badlogic:
That's a great idea. Thanks!

PR #2845, opened 2025-05-12
Fix bug when fetching atlas texture with loader path set.

strider2k:
When setting a path via scene.load.setPath and loading a spine atlas, the texture(s) urls should not have path be in it's url otherwise the loader will doubly prepend the path.

EG:
this.load.setPath("assets/");
this.load.spineBinary("spineboy-data", "spineboy-pro.skel");
this.load.spineAtlas("spineboy-atlas", "spineboy-pma.atlas");

When atlas is fetched, the loader fetch to texture spineboy-pma.png will incorrect have src assets/assets/spineboy-pma.png.


davidetan:
Good catch!
I had to apply it to `phaser-v4` as well, and I did it slightly differently, so I didn’t merge it directly.
I’ll release it in a couple of days along with some other things.

PR #2857, opened 2025-05-20
[c] fix heap corruption while using RGB2Timeline entities

rbolgov:


badlogic:
Cheers!

badlogic:
Had to revert this, the indices are actually wrong.

PR #2860, opened 2025-05-23
[csharp] Synchronized docs with reference implementation

badlogic:
Possible invalid code change in MeshAttachmen.cs

```patch
-                               if (parentMesh != null) throw new ArgumentException("Use newLinkedMesh to copy a linked mesh.");
+                               if (other.parentMesh != null) throw new ArgumentException("Use newLinkedMesh to copy a linked mesh.");
```


HaraldCsaszar:
sorry, not quite there yet, but cool in general!

PR #2874, opened 2025-06-20
Add Godot Spine Skin init method to binds

llwantedl:
* [+] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).

I think that this is a crucial fix for complex games.
During development of my game i couldn't create dynamic skins. Because just having SpineSkin.new() does not initialise the skin, and all other alternatives were just useless as they reference some skin in original spine object one way or another. I was able to compile the GODOT extension with this change, and it is working just fine as i would want it to be.

### DIFF
--- a/spine-godot/spine_godot/SpineSkin.cpp
+++ b/spine-godot/spine_godot/SpineSkin.cpp
@@ -45,6 +45,7 @@ void SpineSkin::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_attachments"), &SpineSkin::get_attachments);
 	ClassDB::bind_method(D_METHOD("get_bones"), &SpineSkin::get_bones);
 	ClassDB::bind_method(D_METHOD("get_constraints"), &SpineSkin::get_constraints);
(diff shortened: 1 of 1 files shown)


llwantedl:
my usecase

```
@onready var spine: SpineSprite = $Sprite

func apply_equipment_to_spine():
	var skeleton = spine.get_skeleton()
	var data = skeleton.get_data()

	var base_skin = data.find_skin(default_skin)
	var combined = SpineSkin.new()
	combined.init("combined", spine)
	combined.copy_skin(base_skin)
	
	# HERE ADDING MORE SLOT-BASED COMBINED SKINS TO THE combined VARIABLE WITH combined.add_skin(item)
        
	skeleton.set_skin(combined)
	skeleton.set_to_setup_pose()
	spine.get_animation_state().apply(skeleton)
```

And without init here combined skin is just empty. Please let me know if there is another way.

badlogic:
Thanks!

PR #2875, opened 2025-06-24
 [ts][pixi-v8] Fix copy method, use structuredClone instead of manually creating array copies

tsingwong:
When I use spine-pixi-v8, there is an error.

CleanShot 2025-06-24 at 12 48 22@2x

some code like this:
```
        const customMesh = slot.getAttachment().copy()
        customMesh.region = textureRegion;
        customMesh.updateRegion();
        slot.setAttachment(customMesh);
```

After some time, I got the reason.

https://github.com/EsotericSoftware/spine-runtimes/blob/5967b3c3f0e568d8278c71c61b9f29d2bd1af876/spine-ts/spine-core/src/attachments/MeshAttachment.ts#L177-L182

CleanShot 2025-06-24 at 12 53 44@2x

https://github.com/EsotericSoftware/spine-runtimes/blob/5967b3c3f0e568d8278c71c61b9f29d2bd1af876/spine-ts/spine-pixi-v8/src/Spine.ts#L656

The reason for this problem is that the original type of UVS was Float32Array, but after the copy method, it became Array and lost the buffer property.

---
So, it can be implemented using [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) of the web standard.
If some browser do not support `structuredClone`, they can use some polyfill.
eg:
```
import { defineConfig } from 'vite'
import react from '@vitejs/plugin-react-swc'
import legacy from '@vitejs/plugin-legacy'


// https://vite.dev/config/
export default defineConfig({
  plugins: [react(),
  legacy({
    targets: ['chrome>=62', 'defaults', 'not IE 11'],
    modernTargets: ["chrome>=62"],
   modernPolyfills:['web.structured-clone']
  }),
  ]
})
```

### DIFF
--- a/spine-ts/spine-core/src/attachments/Attachment.ts
+++ b/spine-ts/spine-core/src/attachments/Attachment.ts
@@ -144,14 +144,12 @@ export abstract class VertexAttachment extends Attachment {
 	/** Does not copy id (generated) or name (set on construction). **/
 	copyTo (attachment: VertexAttachment) {
 		if (this.bones) {
-			attachment.bones = new Array<number>(this.bones.length);
-			Utils.arrayCopy(this.bones, 0, attachment.bones, 0, this.bones.length);
+			attachment.bones = structuredClone(this.bones)
 		} else
 			attachment.bones = null;
 
 		if (this.vertices) {
-			attachment.vertices = Utils.newFloatArray(this.vertices.length);
-			Utils.arrayCopy(this.vertices, 0, attachment.vertices, 0, this.vertices.length);
+			attachment.vertices = structuredClone(this.vertices)
 		}
 
 		attachment.worldVerticesLength = this.worldVerticesLength;
--- a/spine-ts/spine-core/src/attachments/MeshAttachment.ts
+++ b/spine-ts/spine-core/src/attachments/MeshAttachment.ts
@@ -174,20 +174,16 @@ export class MeshAttachment extends VertexAttachment implements HasTextureRegion
 		copy.color.setFromColor(this.color);
 
 		this.copyTo(copy);
-		copy.regionUVs = new Array<number>(this.regionUVs.length);
-		Utils.arrayCopy(this.regionUVs, 0, copy.regionUVs, 0, this.regionUVs.length);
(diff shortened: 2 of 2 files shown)


davidetan:
I cannot reproduce the issue.

When you say:
> The reason for this problem is that the original type of UVS was Float32Array, but after the copy method, it became Array and lost the buffer property.

That shouldn't happen. As you can see from the code of that function from PixiJS repo:
https://github.com/pixijs/pixijs/blob/375a65d0c6fee39620602aee003df1be4959e003/src/rendering/renderers/shared/buffer/utils/fastCopy.ts#L9-L30

The function does not create new elements, but uses the two given in input.

Can you share some code so that I'm able to reproduce it and determine why the `uvs` array change its type?

tsingwong:
I think this is the reason.

https://github.com/EsotericSoftware/spine-runtimes/blob/5967b3c3f0e568d8278c71c61b9f29d2bd1af876/spine-ts/spine-core/src/attachments/MeshAttachment.ts#L177-L182


demo is here: https://github.com/tsingwong/spine-pixi-error-demo
Click the button "change clothes"

Then 
CleanShot 2025-06-24 at 15 28 41@2x

davidetan:
Sorry, I misread your opening post. The issue is pretty straightforward then.

`copy.uvs = new Array<number>(this.uvs.length);` should be: `copy.uvs = Utils.newFloatArray(this.uvs.length);`.

I'll fix it and make a new release.

tsingwong:
> Sorry, I misread your opening post. The issue is pretty straightforward then.
> 
> `copy.uvs = new Array<number>(this.uvs.length);` should be: `copy.uvs = Utils.newFloatArray(this.uvs.length);`.
> 
> I'll fix it and make a new release.

I still recommend using the web standard `structuredClone` method.

davidetan:
We'll stay with what we already have now, and if we'll have issues in the future we'll take in consideration your proposal.
I've just released the fixed version in 4.2.84.

tsingwong:
I just tried it.
You are right, no problem now.

PR #2876, opened 2025-06-25
[ts][pixi] Allow loading from Object URL

### LINKED ISSUE #2744: [ts][pixi-v8][pixi-v7]  Support for Base64-Encoded Assets
The `spine-pixi` runtime does not directly support base64-encoded assets.  

Currently, to make Spine objects work with base64 assets, we need to replicate the functionality of our custom atlas and skeleton loaders.  

This [forum post](https://esotericsoftware.com/forum/d/28081-how-to-load-embedded-skel-and-atlas-files-in-base64-in-pixijs--spine/6) demonstrates the current workaround. As shown, the process is not straightforward, requiring users to manually add assets to the Pixi cache, and to manually loop over all atlas pages and assign the corresponding `PixiSpine` textures.  

This issue proposes enhancing our custom loaders to support base64-encoded assets natively, streamlining the process for users.

krtr:
* [ ] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).

Fixes #2744 

Fixes case when you want to load spine through HTML input element 

Seems like we were mostly there, the only missing stuff was:

* There was no `name` set for loader 
    * Needed for `loadParser` param handing - idomatic way of forcing loaders on assets
    * More info https://pixijs.com/8.x/guides/components/assets/resolver#using-unresolved-assets 
* The `testParse` method held assumption that extension is present in url


davidetan:
The goal of https://github.com/EsotericSoftware/spine-runtimes/issues/2744 is to use base64 url in order to embed the assets within the webpage, a requirement for certain small applications like banner ads.
The example pointed in the issue is just a temporary workaround. The idea is to allow users to pass assets directly in base64 form to the Spine factory `from`.

I'm sorry, but I don't see the point of this PR. Here you are:
1) Loading the assets using fetch from a remote URL
2) Making the assets a blob
3) Creating an object urls from the blobs
4) Passing the object urls to the assets loader explicitly saying to load that assets using the spine one

Can't you just pass the remote URL to the Spine factory `from`?

krtr:
> The goal of #2744 is to use base64 url in order to embed the assets within the webpage, a requirement for certain small applications like banner ads.

Yes you are right.

My issue is a different one, I'm creating pixi/spine based editor which requires user to Drag&Drop spine export files.
Browsers expose those D&Ded files through Blobs.

Why I'm referencing the issue which uses base64 instead of Blobs? 
Blob/files and base64 encoded assets are intechangable with https://developer.mozilla.org/en-US/docs/Web/API/FileReader

davidetan:
Oh, I misinterpreted your goal, but it was quite clear from the PR title.
We need to test the support for binary skeleton files too.

davidetan:
I eventually needed to do the same thing for the skeleton loader. 
I added an `inline-loading.html` example where it's demonstrated asset loading from both base64 and blob.
Thanks for sharing your solution :)

It will be shortly available in `4.2.86`.

luke-thompson-vgw:
Thanks for adding this in. I just integrated it into a project in the last couple of weeks and had it working nicely.

Pixi has just released a new version (8.12.0) changing the from `loadParser` to `parser` and using an `id` property instead of `name`.

When using Pixi 8.12.0 I get deprecation warnings in the console. If I change to use `parser: 'spineTextureAtlasLoader'` instead of `loadParser: 'spineTextureAtlasLoader'` the spine loader doesn't work correctly.

I'm assuming that the required changes would be adding an `id` for the loader and `parser` param handling but am not really sure?

https://github.com/pixijs/pixijs/pull/11557

davidetan:
@luke-thompson-vgw Thanks for informing us. I've updated `4.2` and `4.3-beta` accordingly, and just released `4.2.89` supporting the `parser` param.

PR #2901, opened 2025-08-03
Fix infinite loop when reading json inherit timeline.

MSIsunny:
I found that there is an infinite loop when reading an inherit timeline with multiple keyframes in spine-c runtime;


badlogic:
Ouch, good find. Cheers!

PR #2906, opened 2025-08-13
[ios][4.3-beta] expose explict module for c & cpp

pbk20191:

This Pull Request goal is to add ability to access cpp interface directly using same package for swift package manager system.

- standard user can import spine-c based api without c warning
- dedicated user can import cpp interface with explict import and enabling cpp interp
- objective-c user can import cpp using objective-c++ without issue

there is 3 possibility in user side.
1. swift with  objective-c/C interp mode (default)
2. swift with Cpp interp mode (not default)
3. swift with objective-c++ (this is different from 2 for cpp only case)
 - cpp interface imported by swift interp mode can not be passed back to objective-c/objective-c++ 

swift with cpp interp mode user
```swift
import SpineC.cpp

func dummy() {
    var renderer = spine.SkeletonRenderer()
    
}
import SpineC

func dummyBridged(_ ptr:spine_skeleton_renderer) {
   var renderer = UnsafeMutablePointer<spine.SkeletonRenderer>(OpaquePointer(ptr))

}
```
swift package user can enable their cpp interp mode on their own
```swift
        .target(
            name: "FOO2",
            dependencies: [
                .product(name: "SpineC", package: "spine-runtimes")
            ],
            path: "spine-ios/Sources/Foo2",
            swiftSettings: [
                .interoperabilityMode(.Cxx)
            ]
        ),
```
Xcode project setting coresponds to this is `SWIFT_OBJC_INTEROP_MODE`

objective-c++ user
```objective-c++
/// some .mm file not .m file
#include <spine/spine.h> // this for standard header include, standard for objective-c++
@import SpineC.cpp // this is for clang module import rule, not standard for objective-c++
#import <Foundation/Foundation.h>

void fooo(void) {
    spine::SpineExtension::getInstance();
    
    NSString* foo = @"";
}

```

if the user import cpp header from objective-c file, user will get error named
```
Module 'SpineC.cpp' requires feature 'cplusplus11'
```

if the user tries to import `SpineC.cpp` from swift file, without cpp interp mode enable from their side
```
Module 'SpineC.cpp' requires feature 'cplusplus11'
No such module 'SpineC.cpp'
```

My project highly depends on mixed objective-c/c++ & swift. So ability to directly access spine-cpp interface is really appreciated. ( customized spine::TextureLoader, cpp callback from spine::AnimationState ... etc)


badlogic:
Awesome, thanks!

PR #2913, opened 2025-08-28
[ts][phaser] Update phaser version

GedasTM:
* [ + ] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).


davidetan:
I don't think there's any particular need to upgrade the phaser library right now, unless I am missing something.
It's a peer dependency, so you are free to install `rc5` in your project.

We'll upgrade it when there's a stable release or if necessary, like a breaking change in a new rc.

PR #2917, opened 2025-08-29
[haxe] Remove superfluous OpenFL/lime imports.

eric-hederstedt-hacksaw:
This removes some superfluous OpenFL/lime imports.


davidetan:
Thanks for this. We actually did this changes in the `4.3-beta` branch.
Unfortunately we cannot do the change on `Skeleton` in `4.2` because it would be a breaking change. Users of `getBounds` would have a too simplified `Rectangle` as return value.
I removee the others two import (as we did in `4.3-beta`) in the commit above.

Issue #2930, opened 2025-09-19
[godot] Build fails on Windows when following setup instructions with Godot 4.5


naxa1ka:
### Steps to Reproduce

1. Clone the repo (tested with spine-runtimes 4.2 branch).

   ```bash
   git clone https://github.com/EsotericSoftware/spine-runtimes.git
   cd spine-runtimes/spine-godot
   ```
2. Run setup for Godot 4.5:

   ```bash
   ./build/setup.sh 4.5-stable false
   ```
3. Build the editor:

   ```bash
   ./build/build-v4.sh
   ```

### Expected

Editor for Windows builds successfully.

### Actual

Build fails. See attached log.

### Environment

* **OS:** Windows 10/11 (x64)
* **Godot version:** 4.5-stable
* **Spine runtimes branch:** 4.2

### Logs

[build.log](https://github.com/user-attachments/files/22425074/build.log)

Stvchm9703:
same, I have been try to manually change the code in local build, and seem to be not too stable;
in Godot editor 4.5, they try to implement and change the `String::parse_utf8` to `String::append_utf8`,
while the Godot_cpp (for gdextension), they keep the `String::parse_utf8`.
but even I success build, it will crash when the scene adding to the Node2D; 
maybe wait for godot side to united their code library to fix, so that spine-godot can adapt the changes

---

updated : 2025-09-21 

kind of working, just make sure follow the instruction mention in spine-godot (clean the .nuget etc) 

<img width="1916" height="1340" alt="Image" src="https://github.com/user-attachments/assets/8d1cfe0f-bfeb-4c72-aa97-d1b195b13d8b" />

lukeeingram:
Just giving a quick update: I'm currently working on this issue and will try to have it done soon.

<img width="1900" height="1032" alt="Image" src="https://github.com/user-attachments/assets/f88621ef-7905-45c5-8eca-bcd04e330067" />

lukeeingram:
@naxa1ka @Stvchm9703 I'm closing this issue as my PR adding support for Godot 4.5 was merged onto the `4.3-beta` branch of the Spine Runtimes. If you clone/switch to the `4.3-beta` branch and run the normal setup steps for building spine-godot 4.5, you'll be able to build the spine-godot module or GDExtension.

Once the 4.3-beta branch of the Spine Runtimes exits beta, we'll put links to download the CI builds of spine-godot 4.5-stable in the spine-godot documentation.

PR #2931, opened 2025-09-21
fix: parse_utf8 => utf8 & godot4.5 paths

o-tokar:
I think I've encountered the same problem. [2930](https://github.com/EsotericSoftware/spine-runtimes/issues/2930).
Tested on linux with the GCC 11.4.0


badlogic:
Thanks, but this would break spine-godot for Godot versions < 4.5. We are working on an upgrade to Godot 4.5.

PR #2947, opened 2025-10-14
[android]optimize skeleton drawable load, Avoid app crashes caused by load error

venusic:
Prevent crashes during skeleton loading by ensuring UI initialization runs on the main thread and by catching/logging load failures.

**Key changes:**
Added @MainThread setSkeletonDrawable(...) to centralize init and frame scheduling.
Made loadFrom(AndroidSkeletonDrawableLoader) public, async, with try/catch; posts setSkeletonDrawable to main thread; logs errors on failure.
loadFromDrawable(...) now posts to the main thread.

**Impact:** 
Avoids crashes on load errors; enforces UI-thread correctness; no breaking changes.


badlogic:
Good catch, thanks!

PR #2950, opened 2025-10-17
[ts][pixi-v7][pixi-v8] 4.3 make spine init more universal

Wiking-2-0:
Previously, the `from` method handled both data preparation and instance creation. This made it difficult to extend the `Spine` class or reuse its initialization logic.
With this change, all data preparation (asset loading, parsing, and caching) is now handled inside `getSpineInitData`, while `from` only creates a new instance using the returned data.

### DIFF
--- a/spine-ts/spine-pixi-v7/example/bounds.html
+++ b/spine-ts/spine-pixi-v7/example/bounds.html
@@ -29,21 +29,21 @@
 
 
         // Create the spine display object
-        const spineboy1 = spine.Spine.from({skeleton: "spineboyData", atlas: "spineboyAtlas", scale: .2 });
+        const spineboy1 = new spine.Spine({skeleton: "spineboyData", atlas: "spineboyAtlas", scale: .2 });
 
-        const spineboy2 = spine.Spine.from({skeleton: "spineboyData", atlas: "spineboyAtlas", scale: .2,
+        const spineboy2 = new spine.Spine({skeleton: "spineboyData", atlas: "spineboyAtlas", scale: .2,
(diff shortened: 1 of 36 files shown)


davidetan:
As I told you in the forum, the `from` initialization isn't the default Spine initialization, but it's just a convenient way of initializing the Spine object. For example, rather than using the `AtlasAttachmentLoader` you might want to use your AttachmentLoader.

In any case, I didn't like the `from` method even if it is idiomatic to PixiJS. I like having a static method to get the convenient initialization configuration. Keeping `from` would be just for retrocompatibility, but even if in `4.3` we can make breaking changes, I prefer to just mark it as deprecated for now.

Moreover, this allows us to simplify the constructor and use interchangably the `SpineFromOptions` and the `SpineOptions`. If the former is passed, the constructor will call directly the static method to create the `SpineOptions`.

I pushed these changes (and renamed `getSpineInitData` to `createOptions`, the `Spine` is implicit). Let me know if you think we need to make any other change before merging this.

Wiking-2-0:
Looks perfect!
I really like the new initialization approach.

One question.
Shouldn’t this line be updated now that we removed the `instanceof SkeletonData` check above?
`const skeletonData = options instanceof SkeletonData ? options : options.skeletonData;`

Line 396: https://github.com/EsotericSoftware/spine-runtimes/pull/2950/commits/512c5a12d5d01261a6feb443816541563029bd7a#diff-f0503b7467567913e50ac50c7a37fb0298bd08c925d32c465d29897cafea7b23L398

davidetan:
Yup, I missed that. Fixed a slightly refactored constructors. I'm going to merge it. Good job!

PR #2955, opened 2025-10-19
[ts] Replace Array constructors with dynamic initialization for packed number arrays

koteelok:
In some places, you use `new Array(size)` to make number arrays. But JS fills the array with empty space and V8 will treat them as [`HOLEY_SMI_ELEMENTS` and `HOLEY_DOUBLE_ELEMENTS` instead of `PACKED_SMI_ELEMENTS` and `PACKED_DOUBLE_ELEMENTS`](https://v8.dev/blog/elements-kinds).

I grepped through codebase and fixed this in some places, but I don’t think it will result in any noticeable speed-up. I just "scratched the itch"

Feel free to reject, if you don't care :P


davidetan:
I agree that it won’t provide any noticeable  benefit, but it’s less verbose code and matches our new linter better. I checked to ensure it wouldn’t cause any issues, and I can confirm all changes are safe.

I had to merge the 4.3 branch into yours since there were some conflicts with my changes.

To accept your merge, you need to agree to our terms.

Could you please send a message with [this text](https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/.github/pull_request_template.md) and check the box? Then I can merge. Thanks! :)

koteelok:

davidetan:
Thanks! Merged :)

PR #2961, opened 2025-10-27
[ts][phaser-v3] fix atlas page url duplication when using file:// + baseURL

aomsir:
#### 🧩 Summary
When loading Spine atlases under the **`file://` protocol** with a `baseURL` set, atlas page PNGs still produce double-prefixed URLs like:

```
file:///Users/.../spine--demo.../file:///Users/.../assets/spine/spineboy-pma.png
```

All other assets load correctly — only Spine atlas **page images** (`.png` listed in the `.atlas` file) are affected.

This happens because the plugin uses `file.src` to determine the atlas base directory. In this case `file.src` already includes the `baseURL`, so the loader prepends `baseURL` again.

---

#### 🧠 Relation to Phaser issue [#6642](https://github.com/phaserjs/phaser/issues/6642)
Phaser core loader fixed the same issue in [#6642](https://github.com/phaserjs/phaser/issues/6642), replacing `file.src` with `file.url` inside `GetURL()` to avoid double baseURL concatenation.

However, this Spine plugin (`spine-phaser-v3`) still uses `file.src` in `onFileComplete()`, so the duplication persists when running under `file://` with `setBaseURL()`.

---

#### 🧾 Changes
```diff
// src/SpinePlugin.ts (method onFileComplete)

- let basePath = (file.src.match(/^.*\//) ?? "").toString();
+ let basePath = (file.url.match(/^.*\//) ?? "").toString();
```

This aligns the plugin with the Phaser core fix from #6642 and ensures consistent behavior across protocols.

---

#### ✅ Results

| Environment       | Case                         | `file.url`                                  | `file.src`                                          | Outcome |
|-------------------|------------------------------|---------------------------------------------|-----------------------------------------------------|---------|
| HTTP / HTTPS      | Any asset                    | `assets/bg.png`                             | `http://localhost/.../assets/bg.png`                | ✅ OK   |
| `file://` (before)| Spine atlas page `.png`      | `file:///Users/.../assets/spineboy-pma.png` | `file:///base/file:///Users/.../spineboy-pma.png`   | ❌ BAD  |
| `file://` (after) | Spine atlas page `.png`      | `file:///Users/.../assets/spineboy-pma.png` | `file:///Users/.../assets/spineboy-pma.png`         | ✅ OK   |

---

#### 📝 Notes
- No impact on HTTP(S) or other environments.
- Fixes `file:// + baseURL` compatibility for atlas page PNGs.
- Brings plugin behavior in line with Phaser core loader fix (#6642).
- Verified on Phaser 3.90.0, macOS + Chrome.

---

### DIFF
--- a/spine-ts/spine-phaser-v3/src/SpinePlugin.ts
+++ b/spine-ts/spine-phaser-v3/src/SpinePlugin.ts
@@ -374,7 +374,9 @@ class SpineAtlasFile extends Phaser.Loader.MultiFile {
 					}
 				}
 
-				let basePath = (file.url.match(/^.*\//) ?? "").toString();
+				let fileUrl = file.url;
+				if (typeof fileUrl === "object") fileUrl = file.src;
+				let basePath = (fileUrl.match(/^.*\//) ?? "").toString();
 				if (this.loader.path && this.loader.path.length > 0 && basePath.startsWith(this.loader.path))
 					basePath = basePath.slice(this.loader.path.length);
 
--- a/spine-ts/spine-phaser-v3/src/SpinePlugin.ts
+++ b/spine-ts/spine-phaser-v3/src/SpinePlugin.ts
@@ -374,7 +374,7 @@ class SpineAtlasFile extends Phaser.Loader.MultiFile {
 					}
 				}
 
-				let basePath = (file.src.match(/^.*\//) ?? "").toString();
+				let basePath = (file.url.match(/^.*\//) ?? "").toString();
 				if (this.loader.path && this.loader.path.length > 0 && basePath.startsWith(this.loader.path))
 					basePath = basePath.slice(this.loader.path.length);
 
--- a/spine-ts/spine-phaser-v4/src/SpinePlugin.ts
+++ b/spine-ts/spine-phaser-v4/src/SpinePlugin.ts
@@ -366,7 +366,9 @@ class SpineAtlasFile extends Phaser.Loader.MultiFile {
 					}
 				}
 
-				let basePath = (file.src.match(/^.*\//) ?? "").toString();
+				let fileUrl = file.url;
+				if (typeof fileUrl === "object") fileUrl = file.src;
+				let basePath = (fileUrl.match(/^.*\//) ?? "").toString();
 				if (this.loader.path && this.loader.path.length > 0 && basePath.startsWith(this.loader.path))
 					basePath = basePath.slice(this.loader.path.length);
 


aomsir:
If possible, please consider merging this fix soon and including it in the next **4.2.95** release.  
We rely on running Phaser + Spine under the `file://` protocol for internal tools,  
and this issue currently requires us to patch the plugin source manually before deployment.  
Having the fix officially released would prevent us from maintaining a local fork.  
Thank you very much for your time and for maintaining this excellent plugin!

davidetan:
It seems our official plugin was ported before that fix. We'll apply it here as well, thanks for pointing it out in the Phaser repo.
We just need to reproduce it and verify it doesn’t break other use cases.

aomsir:
> It seems our official plugin was ported before that fix. We'll apply it here as well, thanks for pointing it out in the Phaser repo. We just need to reproduce it and verify it doesn’t break other use cases.

Thanks a lot for the quick response!

Just to share a bit more detail — I’ve tested both approaches:

1. Updating the **Phaser core** to recognize `file://` as an absolute URL (so `GetURL()` doesn’t prepend the baseURL again);
2. Updating this **plugin** to use `file.url` instead of `file.src` when deriving the atlas base path.

Both methods resolve the duplication issue under the `file://` protocol, and everything else continues to work as expected.  
So please feel free to test and verify — I’m confident the fix is safe. ✅

By the way, may I ask roughly when the **4.2.95** release might be planned?  
If it’s possible within the next day or two, that would be amazing — I’m currently preparing a build that depends on this fix.  
Thanks again for your time and support!

davidetan:
I was able to test this, and the issue does occur.

I'm releasing this right after I merge it, don't worry.

---

For future reference for me, to fetch an asset using the `file://` protocol, you need to launch Chrome with the following flags:

```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome \
  --disable-web-security \
  --allow-file-access-from-files \
  --user-data-dir="/tmp/chrome_dev"
```

However, this alone isn’t enough, it only allows fetching a file when the HTML is loaded directly as a file (`--allow-file-access-from-files`). Therefore, we need to modify our basic example like this:

```
...
<script src="../../assets/phaser.js"></script> // download Phaser locally (placed under assets for convenience)
...
...
preload() {
  this.load.spineBinary("spineboy-data", "../../assets/spineboy-pro.skel");
  this.load.spineAtlas("spineboy-atlas", "../../assets/spineboy-pma.atlas");
}
```

(If it doesn’t work, try deleting the `/tmp/chrome_dev` folder and reopening Chrome.)

aomsir:
> I was able to test this, and the issue does occur.
> 
> I'm releasing this right after I merge it, don't worry.
> 
> For future reference for me, to fetch an asset using the `file://` protocol, you need to launch Chrome with the following flags:
> 
> ```
> /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome \
>   --disable-web-security \
>   --allow-file-access-from-files \
>   --user-data-dir="/tmp/chrome_dev"
> ```
> 
> However, this alone isn’t enough, it only allows fetching a file when the HTML is loaded directly as a file (`--allow-file-access-from-files`). Therefore, we need to modify our basic example like this:
> 
> ```
> ...
> <script src="../../assets/phaser.js"></script> // download Phaser locally (placed under assets for convenience)
> ...
> ...
> preload() {
>   this.load.spineBinary("spineboy-data", "../../assets/spineboy-pro.skel");
>   this.load.spineAtlas("spineboy-atlas", "../../assets/spineboy-pma.atlas");
> }
> ```
> 
> (If it doesn’t work, try deleting the `/tmp/chrome_dev` folder and reopening Chrome.)

That’s great to hear — thank you so much for verifying and for the super fast turnaround!  
Really appreciate the detailed reproduction notes as well.  
Looking forward to the 4.2.95 release

davidetan:
I had to make a small change due to the returned type being `string | object`.
I did the same on `phaser-v4`.
Now released in `4.2.95`. 

Thanks for your contribution!

PR #2970, opened 2025-11-07
[iOS][4.3-beta] move `SpineiOS` target behind `SpineiOSWrapper`

pbk20191:

There is no good way to conditionally add Swift package Product in single project Target  (Multiplatform Target). 

previous 4.2 does not falls to build failure when building Multiplatform Target even if Spine is not supported (macOS)
It is just gated `SpineModule` https://github.com/EsotericSoftware/spine-runtimes/blob/4.2/Package.swift#L24-L37

which drops actual module `Spine`. So there is no issue.

But for 4.3 beta gating module is gone, and now it is not possible to build Multiplatform Target, since xcode always tries to compile spine where it is not supported (like trying to compile SpineiOS for macOS instead of just dropping the target, even if the user guarded with macro)

So I added the gating module again to fix this problem

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -25,10 +25,17 @@ let package = Package(
         ),
         .library(
             name: "SpineiOS",
-            targets: ["SpineiOS"]
+            targets: ["SpineiOSWrapper"]
         ),
     ],
     targets: [
+        .target(
+            name: "SpineiOSWrapper",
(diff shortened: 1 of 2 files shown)


badlogic:
Really thankful for your contribtions! The iOS/macOS toolchain is very confusing to me :)

PR #2975, opened 2025-11-12
[ios] backport from 4.3-beta: expose explict module for c & cpp

pbk20191:
- this is the backport version of  #2906 to spine-4.2
- In addition, expose spine-cpp module directly like spine 4.3-beta
- standard user can import spine-c based api as usual
- dedicated swift/cpp user can import cpp interface with explict import and enabling cpp interp
- objective-c user can import cpp using objective-c++


badlogic:
Cheers!

PR #2995, opened 2025-12-11
[ts] pma on upload for textures

davidetan:
This PR adds support across all runtimes for premultiplying non-PMA textures on upload.

The advantages are:

* All PMA benefits also apply to non-PMA textures
* Compressed WEBP can use PMA features (since non-PMA textures are premultiplied on upload)
* The code is simplified:
  * Users no longer need to pass any PMA settings
  * No need to branch color calculations based on PMA, since everything is PMA (most libraries/runtimes use PMA)

If merged, we need to update all TS-related documentation by removing the PMA settings and mentioning that non-PMA and PMA textures can be used interchangeably.


spine-ts/spine-core/src/AssetManagerBase.ts
```diff
@@ -35,20 +35,19 @@ type AssetData = (Uint8Array | string | Texture | TextureAtlas | object) & Parti
...
 	private toLoad = 0;
 	private loaded = 0;
+	private texturePmaInfo: Record<string, boolean> = {};
 
-	constructor (textureLoader: (image: HTMLImageElement | ImageBitmap) => Texture, pathPrefix: string = "", downloader = new Downloader(), cache = new AssetCache()) {
-		this.textureLoader = textureLoader;
-		this.pathPrefix = pathPrefix;
-		this.downloader = downloader;
-		this.cache = cache;
+	constructor (
```
badlogic, line 46: Davide, code janitor.

PR #3011, opened 2026-01-17
[ts][pixi-v8] Added example shadow-projection-shader

HarmHoeksema:


davidetan:
Thanks for the PR, but this should probably be sent to Pixi rather than us.
It’s a generic filter, not something specific to Spine.

PR #3037, opened 2026-03-04
[cpp] fixed Memory access out of bounds error caused by unhandled null return of RegionAttachment when loading default skin

driftboat:
When the texture packer is set to ignore blank images and exports a binary animation with empty sprites,
during the loading of the default skin in the skeleton file, the RegionAttachment returns null,
but there is no error handling or early return,
so the parsing continues, leading to misalignment and corruption in subsequent reads.


badlogic:
cheers, backported manually.

PR #3039, opened 2026-03-08
Add overloads for SkeletonRendererCocos2dX methods findBone, findSlot and getAttachment accepting const char * instead of std::string.

chemodansama:
Those methods did not have const char * overloads, though setSkin() method has such overloads.
Also, existing methods accepting std::string were constructing spine::String implicitly which caused it to allocate a copy on the heap. This commit changes it so that spine::String is constructed without allocations/deallocations by passing constructor parameters own == true and tofree == false.


badlogic:
cocos2dx is no longer part of spine-runtimes as of the 4.3 relesae.

PR #3043, opened 2026-03-17
[godot] Resolves #3036, adds GDExtension implementation of animation mixes editor inspector panel

### LINKED ISSUE #3036: [Godot] Editor crashes when adding elements to the animation mixes array in the Skeleton Data Res
Whenever I try to modify the fields, especially the animation mixes array on a Skeleton Data Resource, the Godot Editor will crash immediately.  It  happens when trying to modify existing as well as brand new Skeleten Data Resources. The screenshot is taken from the samples package, where it also happens. 

I'm using the GDExtension for Godot 4.5.1 right now, but I've also encountered this issue on 4.6.1 on Pop OS 24.04 LTS.

To reproduce, just go to the samples project, open a SpineSprite, open the animation mixes array and click on "Add". 

Image

lukeeingram:
This directly resolves the crash with issue #3036 and adds the ability for users to manipulate animation mixes in the Editor's inspector panel.

### DIFF
--- a/spine-godot/spine_godot/SpineCommon.h
+++ b/spine-godot/spine_godot/SpineCommon.h
@@ -32,6 +32,7 @@
 
 #ifdef SPINE_GODOT_EXTENSION
 #include <godot_cpp/core/version.hpp>
+#include <godot_cpp/classes/ref.hpp>
 
 // When running scons with deprecated=no, these are not defined in version.h in Godot 4.5.1
 // but our code for older versions of Godot relies on them.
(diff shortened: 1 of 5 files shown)


badlogic:
thanks @lukeeingram, good job! I ported this to 4.3 and landed it in 185905297. I also added a couple extra guards around null/invalid mixes and fixed a small indexing bug in set_animation_mixes(). Built the macOS GDExtension and verified the animation mix inspector in Godot 4.6.2 manually.

PR #3053, opened 2026-04-03
[ts][pixi-v8] Update slot object container with tint

Hypnotriod:
* [+] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).


davidetan:
I was wondering, what if a user does not want the tint to extend to the slot object?
We should probably add an option to let the user decide if the slot object has to be tinted as the spine object.

Hypnotriod:
Hi. In our case we have the Pixi Text inside the Spine slot object. And we want the colors of that text follow the Spine tint animation. So we have patched the Spine runtime for our needs.
If you are already propagating the slot alpha channel, so why not the rest of the colors?
But it is only a proposal. Maybe there is a better way?

davidetan:
We decided to implement this through an opt-in behavior using a new `followSlotColor` property. Thanks for the idea and your contribution.

Released in `4.2.119` and `4.3.3`.

PR #3056, opened 2026-04-12
fix: support phaser v4 esm runtime

### LINKED ISSUE #3055: [ts][phaser-v4] ESM/Vite integration fails at runtime: `ReferenceError: Phaser is not defined`
# [ts][phaser-v4] ESM/Vite integration fails at runtime: `ReferenceError: Phaser is not defined`

## Runtime versions

- Phaser: `4.0.0`
- `@esotericsoftware/spine-phaser-v4`: `4.2.109`
- Vite: `7.1.x`
- TypeScript: `5.9.x`

## Summary

A clean TypeScript + Vite + Phaser 4 setup using:

```ts
import * as Phaser from "phaser";
import { SpinePlugin } from "@esotericsoftware/spine-phaser-v4";
```

builds successfully, but fails at runtime when the Spine package is evaluated.

Observed error:

```text
Uncaught ReferenceError: Phaser is not defined
    at @esotericsoftware_spine-phaser-v4.js:14771
```

## Expected behavior

The official npm package should work in a normal ESM/Vite TypeScript setup without requiring:

- a global `window.Phaser`
- a custom bootstrap shim
- vendored browser/IIFE runtime files

## Actual behavior

The app compiles and bundles, but the Spine runtime crashes at startup unless I manually provide a browser-global `Phaser` compatibility layer before loading `@esotericsoftware/spine-phaser-v4`.

## Minimal repro

```ts
import * as Phaser from "phaser";
import { SpinePlugin } from "@esotericsoftware/spine-phaser-v4";

class DemoScene extends Phaser.Scene {
  constructor() {
    super("DemoScene");
  }

  preload() {
    this.load.spineJson("spineboy-data", "/path/to/spineboy-pro.json");
    this.load.spineAtlas("spineboy-atlas", "/path/to/spineboy.atlas");
  }

  create() {
    const spineObject = this.add.spine(480, 520, "spineboy-data", "spineboy-atlas");
    spineObject.setScale(2.2);
  }
}

const config: Phaser.Types.Core.GameConfig = {
  type: Phaser.WEBGL,
  parent: "app",
  width: 960,
  height: 720,
  scene: [DemoScene],
  plugins: {
    scene: [
      {
        key: "spine.SpinePlugin",
        plugin: SpinePlugin,
        mapping: "spine",
      },
    ],
  },
};

new Phaser.Game(config);
```

## Repro steps

1. Create a basic Vite + TypeScript project.
2. Install:
   - `phaser@4.0.0`
   - `@esotericsoftware/spine-phaser-v4@4.2.109`
3. Add the code above.
4. Use any valid Spine sample asset set, for example the official Spineboy sample.
5. Run the dev server.
6. Open the page.

## Notes

From inspecting the published package, the runtime appears to contain free global `Phaser` references inside the ESM bundle, instead of relying only on imported Phaser module references. That seems to be why a standard ESM setup fails unless `window.Phaser` is prepared manually first.

The failing references appear to come from the published ESM bundle around:

- `Phaser.GameObjects.Components`
- `Phaser.Class.mixin(...)`

I can provide a zipped repro if helpful.

frederikocmr:
# `[phaser-v4] Fix ESM runtime initialization in Vite/TypeScript projects`

Fixes #3055.

## Summary

This fixes the Phaser 4 ESM runtime path for `spine-phaser-v4` in a plain Vite/TypeScript setup.

Before this change, the package could build successfully but still fail at runtime with:

```text
ReferenceError: Phaser is not defined
```

The problem was that the runtime mixed module-based Phaser usage with free global `Phaser` references.

## Root cause

- `SpinePlugin.ts` already works as a normal Phaser module import.
- `SpineGameObject.ts` still referenced `Phaser` without importing it.
- `mixins.ts` also depended on a free `Phaser` global.
- `mixins.ts` called `Phaser.Class.mixin(...)`, but Phaser's ESM build does not expose `Class`.

That combination breaks a normal setup like:

```ts
import * as Phaser from "phaser";
import { SpinePlugin } from "@esotericsoftware/spine-phaser-v4";
```

## Changes

- Add an explicit Phaser import in `SpineGameObject.ts`.
- Add an explicit Phaser import in `mixins.ts`.
- Replace `Phaser.Class.mixin(...)` with a small local mixin helper.
- Preserve getter/setter descriptors and skip non-configurable properties, matching the expected mixin behavior for Phaser game object components.

## Validation

- Built the `spine-ts` workspace successfully on `4.3-beta`.
- Verified the fix against a minimal Phaser 4 + Vite + TypeScript repro.
- Confirmed the repro now loads and renders correctly without:
  - `window.Phaser`
  - a custom bootstrap shim
  - vendored browser/IIFE runtime files

## Contributor checklist

- [x] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).

### DIFF
--- a/spine-ts/spine-phaser-v4/src/SpineGameObject.ts
+++ b/spine-ts/spine-phaser-v4/src/SpineGameObject.ts
@@ -27,6 +27,7 @@
  * THE SPINE RUNTIMES, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *****************************************************************************/
 
+import * as Phaser from "phaser";
 import {
 	AnimationState,
 	AnimationStateData,
--- a/spine-ts/spine-phaser-v4/src/mixins.ts
+++ b/spine-ts/spine-phaser-v4/src/mixins.ts
@@ -26,6 +26,8 @@ SOFTWARE.
 
 // biome-ignore-all lint: ignore biome for this file
 
+import * as Phaser from "phaser";
+
 const components = (Phaser.GameObjects.Components as any);
 export const ComputedSize = components.ComputedSize;
 export const Depth = components.Depth;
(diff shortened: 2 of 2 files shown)


davidetan:
It wasn't working with an IIFE in my tests.
I don't think we need the "Preserve getter/setter descriptors and skip non-configurable properties, matching the expected mixin behavior for Phaser game object components" part either. The extra complexity handles scenarios that don't apply to our use case. We are copying Phaser component prototypes onto a fresh class.

I used a simpler approach that adds less complexity but should achieve the same result.

PR #3068, opened 2026-04-24
[ts] fix: on some devices, a canvas exceeding the maximum texture size can…

DelTheCode:
… cause rendering distortion.

**Problem phenomenon**: On some foldable phones, due to the large screen width and DPR, the canvas width is set to be excessively large, exceeding the maximum size limit of the texture map, resulting in animation rendering distortion.

**Recurrence conditions**：Galaxy Z Fold7, maxTextureSize=4096, when canvas size lager than 4096, resulting in animation rendering distortion.

**Program Summary**: When the canvas size is larger than the maximum texture size requirement, the maximum texture size is selected as the boundary reference.

### DIFF
--- a/spine-ts/spine-webgl/src/SceneRenderer.ts
+++ b/spine-ts/spine-webgl/src/SceneRenderer.ts
@@ -52,6 +52,7 @@ export class SceneRenderer implements Disposable {
 	canvas: HTMLCanvasElement;
 	camera: OrthoCamera;
 	batcher: PolygonBatcher;
+	maxTextureSize: number;
 	private twoColorTint = false;
 	private batcherShader: Shader;
 	private shapes: ShapeRenderer;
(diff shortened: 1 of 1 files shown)


davidetan:
I agree setting a max is not a bad idea, but could you share more info about your use case.
When that is happening e which is the canvas width in this case?

DelTheCode:
> I agree setting a max is not a bad idea, but could you share more info about your use case. When that is happening e which is the canvas width in this case?

On a foldable test phone with a screen width of 698 and a dpr of 2.75, 
I set the canvas width to the screen width=698 and the height=1843. 
698 * dpr = 1920, 
1843 * dpr = 5068 >  4096(maxTextureSize of this Phone). 
I then observed that the animation was severely stretched. Although this is not common, avoiding this threshold trap will help others avoid similar pitfalls.

davidetan:
We made a fix in 4.2.113. WebGL limits are checked once and on viewport limits.

Issue #2689, opened 2024-11-18
[Godot] Skeleton rendering separation


misaki-eymard:
A `SpineSlotNode` can be used to insert another object between the draw order of a skeleton's slots, but this is not well suited for temporarily inserting objects that were originally moved separately, e.g. two characters normally move separately, and one character is temporarily placed between the other's drawing order for a hug animation. If you try to do this with SpineSlotNode, you have to duplicate the character (one for when they move separately and one for when they move between certain SpineSpriteNodes) because the inserted object must be a child of SpineSpriteNode. It would be convenient if you can temporarily separate the draw order of a skeleton at the front and back, without duplicating the character, and allow another character to be sandwiched in between like the [SkeletonRenderSeparator](https://esotericsoftware.com/spine-unity-utility-components#SkeletonRenderSeparator) component of the spine-unity runtime.

Related forum thread:
https://esotericsoftware.com/forum/d/27253-godot-43-slot-draw-order-manipulation-at-runtime

PR #3091, opened 2026-05-22
[godot] Add render separation support for SpineSprite

GameJawnsInc:
* [✅] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).

## What this does

Implements a Godot equivalent of spine-unity's [`SkeletonRenderSeparator`](https://esotericsoftware.com/spine-unity-utility-components#SkeletonRenderSeparator), addressing #2689.

It lets you draw an unrelated node *between* the front and back halves of a skeleton's draw order — e.g. placing character B between character A's torso and arms during a hug — without duplicating character A.

### New classes (`spine-godot/spine_godot/`)

- **`SpineSlotRangeProxy`** (`Node2D`) — claims a contiguous slot range `[start_slot_name .. end_slot_name]` from a source `SpineSprite` and renders those slots itself. Being an ordinary node, it can be placed anywhere in the scene tree to control where that slice of the skeleton draws relative to other nodes.
- **`SpineSpriteRenderSeparator`** (`Node`) — convenience node mirroring Unity's `SkeletonRenderSeparator`. Given a source sprite and a list of separator slot names, it partitions the skeleton into N+1 contiguous ranges and creates one `SpineSlotRangeProxy` child per range.

### Changes to `SpineSprite`

Adds a proxy registration API (`_register_proxy` / `_unregister_proxy`). Slot meshes claimed by a proxy are hidden on the source so they aren't drawn twice; geometry is still updated every frame so the proxy always mirrors fresh data (including skin/attachment swaps).

### Example

Adds `example-v4/examples/14-render-separator/` — two raptor + spineboy pairs (one flat, one with the spineboy bone-attached to the raptor via a `SpineBoneNode`). Each uses `z_index` on the source so the raptor draws between the spineboy's back and front halves.

## How it works

The source `SpineSprite` remains the single source of truth for vertex data. A proxy doesn't reparent or rebuild meshes — it mirrors the source's per-slot mesh geometry and material into its own canvas items each frame, and the source hides the slots a proxy has claimed. No `spine-cpp` changes; pure Godot integration code.

The proxy mirrors the source during its own `_process`, so it should sit after the source `SpineSprite` in the scene tree (documented on the class). A proxy placed before its source mirrors the previous frame's geometry.

## Verification

Tested on Windows / MSVC:

- Builds clean both as a Godot **module** and as a **GDExtension** (editor, template_debug, template_release).
- `14-render-separator` example renders correctly with the source as a plain sibling and bone-attached to another skeleton — no double-draw.
- Multi-proxy: a `SpineSpriteRenderSeparator` splitting spineboy into 3 ranges composes pixel-identical to the un-separated skeleton.
- No regression: the existing `07-slot-node` example renders unchanged.

## Open questions for maintainers

1. **Naming** — I used `SpineSlotRangeProxy` + `SpineSpriteRenderSeparator`. Prefer mirroring Unity 1:1 (`SkeletonRenderSeparator` / `SkeletonPartsRenderer`)?
2. **Material override** — should `SpineSlotRangeProxy` allow overriding the per-slot blend-mode material, as Unity's `SkeletonPartsRenderer` does?
3. **`SpineSlotNode` children** — if a `SpineSlotNode` is bound to a slot inside a proxy's range, should it render under the source or the proxy? Currently it stays with the source.
4. **Overlapping ranges** — two proxies claiming the same slot currently both draw it (a `WARN_PRINT` is emitted). Is "warn" sufficient, or should the overlap be assigned to exactly one proxy?

---

*Implemented with the assistance of [Claude Code](https://www.anthropic.com/claude-code).*


badlogic:
Screenshot 2026-06-05 at 11 45 33

appreciate the effort, but i'd rather not merge this.

badlogic:
it's especially weird, because the godot runtime already supports this via slot nodes, which is vastly superior to the renderer separator design.

GameJawnsInc:
thanks for taking a look, i understand the hesitation but this does solve a real problem

i couldn't get the current behavior of the "spineboy riding raptor bone" (14-render-separator) example scene using SpineSlotNode. 

if you want the raptor to show in-between spineboy's front arm and leg and the rest of his body, the raptor would need to be a child of spineboy's SpineSlotNode.

so basically the proxy solves the disconnect between slot rendering order and Godot SceneTree order. i attached a GIF showing how the proxy lets you control draw order of a SpineSprite that is already the child of a SpineBoneNode.

render-separator

my example is messy visually, I didn't want to touch spineboy's animations. but with this setup, you could create "Riding" poses on the spineboy rig itself, and you wouldn't have to bake his animations into the raptor skeleton. he can just ride the raptor's SpineBoneNode instead and use a proxy to put his back half behind the raptor.

it also lets you interlace two SpineSprites together (i.e for a hug), which has the same SceneTree order issue

PR #3125, opened 2026-06-16
[ts][spine-canvas] Apply blend modes, support ClippingAttachment, trim vertex data

obiot:

> Signed CLA is being printed and emailed to contact@esotericsoftware.com.

This patch was developed by the [melonJS](https://melonjs.org) team (as part of the canvas fallback rendering mode we also support) while building [`@melonjs/spine-plugin`](https://www.npmjs.com/package/@melonjs/spine-plugin) — our official Spine integration. We hit each of the bugs while working in the plugin's canvas path, and the fixes are verified against the official 4.3 example skeletons listed below.

## Summary

Three changes to `spine-canvas/src/SkeletonRenderer.ts` — two bringing it to feature parity with `spine-webgl`, plus a vertex-buffer cleanup:

- **Per-slot blend modes** (`additive` / `multiply` / `screen`) — read from `slot.data.blendMode` but never applied to the canvas context today. In `drawTriangles` the value was read into a local tracker and immediately discarded; in `drawImages` there was no blend-mode handling at all. Every other official spine-ts renderer (`spine-webgl`, `spine-pixi-v8`, `spine-threejs`, `spine-phaser-v4`, `spine-player`) handles this. *Visible on `diamond-pro` (the entire mesh-based shine animation renders as a flat chalky outline — the 15 additive slots that drive it draw as opaque overlays) and on `spineboy-pro` (the additive rim-lights on goggles, visor, and jetpack flame draw as opaque decals instead of glows — and `spineboy-pro` is the character `spine-canvas/example/index.html` loads).*

- **`ClippingAttachment` slot-range clipping** — the file didn't import `ClippingAttachment` or `SkeletonClipping`. Skeletons with a clipping attachment rendered the masked geometry outside the clip polygon. (The single `ctx.clip()` in `drawTriangle` is the per-triangle texture-mapping trick, unrelated to Spine's clipping feature.) Same five renderers above all support it. *Visible on `coin-pro` (the additive `shine` slot is clipped to the disc silhouette in the editor; under the current canvas runtime the shine extends past the rim as a vertical bar outside the coin shape) and on `spineboy-pro` (the visor-glass `ClippingAttachment` is ignored, so the lens overlay spills past the goggle frame onto his face).*

- **Vertex-buffer cleanup — positions only (`VERTEX_SIZE` 8 → 2).** The buffer interleaved 4 RGBA floats and 2 UV floats per vertex alongside the position. Canvas2D has no per-pixel vertex-attribute interpolation: `drawTriangle` does `ctx.drawImage(img, 0, 0)` with a uniform per-slot `globalAlpha` — the per-vertex color was computed each frame and never read. UVs come from `sequence.getUVs(sequenceIndex)`, a spine-cached array that's stable per sequence frame, so interleaving them into the per-frame vertex buffer was redundant. Dropping both collapses `computeRegionVertices` / `computeMeshVertices` into a single `computeWorldVertices(..., stride: 2)` call and removes ~50 lines of helper code. For the example `spine-canvas/example/index.html` loads (`spineboy-pro` — 12 mesh + 66 region attachments, ~586 vertices per frame) the per-frame savings are below any measurable threshold; on a crowded scene (many skeletons, or dense character meshes with 1500+ vertices) they scale linearly. Behavior is byte-identical — the dropped writes were never read.

## Affected example skeletons (`spine-ts/assets/`)

Surveyed all 4.3 example skeletons — these render incorrectly under the current canvas runtime:

| Skeleton | Blend slots | Clipping |
|---|---|---|
| **`spineboy-pro`** (the character `spine-canvas/example/index.html` loads) | 25 × additive | ✅ |
| `spineboy-ess` | 1 × additive | |
| `coin-pro` | 2 × additive | ✅ |
| `tank-pro` | 8 × multiply + 2 × additive | ✅ |
| `diamond-pro` | 15 × additive | |
| `powerup-pro` | 10 × additive | |
| `celestial-circus-pro` | 8 × multiply | |
| `snowglobe-pro` | 1 × multiply | |
| `chibi-stickers` | 1 × multiply | |

## Verification

Rebuilt locally and verified `spineboy-pro` (meshes + additive + clipping) and `celestial-circus-pro` (47 mesh + 35 region + 8 multiply, PMA atlas) render correctly via the official `spine-canvas/example` against the patched bundle. Region/mesh paths, sequence-driven attachments, and clip ranges all exercised.

### DIFF
--- a/spine-ts/spine-canvas/src/SkeletonRenderer.ts
+++ b/spine-ts/spine-canvas/src/SkeletonRenderer.ts
@@ -27,21 +27,51 @@
  * THE SPINE RUNTIMES, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *****************************************************************************/
 
-import { type BlendMode, Color, MeshAttachment, type NumberArrayLike, RegionAttachment, type Skeleton, type Slot, type TextureRegion, Utils } from "@esotericsoftware/spine-core";
+import { BlendMode, ClippingAttachment, Color, MeshAttachment, type NumberArrayLike, RegionAttachment, type Skeleton, SkeletonClipping, type Slot, type TextureRegion, Utils } from "@esotericsoftware/spine-core";
 import type { CanvasTexture } from "./CanvasTexture.js";
 
 const worldVertices = Utils.newFloatArray(8);
 
+/** Map a Spine `BlendMode` to the matching Canvas2D `globalCompositeOperation`.
+ * The four Spine blend modes have direct canvas equivalents — "lighter" reproduces
+ * additive (`GL_ONE, GL_ONE`) blending on canvas2d.
+ */
+function blendModeToComposite (blendMode: BlendMode): GlobalCompositeOperation {
+	switch (blendMode) {
+		case BlendMode.Additive: return "lighter";
+		case BlendMode.Multiply: return "multiply";
+		case BlendMode.Screen:   return "screen";
+		default:                 return "source-over"; // Normal
+	}
+}
+
 export class SkeletonRenderer {
 	static QUAD_TRIANGLES = [0, 1, 2, 2, 3, 0];
-	static VERTEX_SIZE = 2 + 2 + 4;
+	// Positions-only world-vertex buffer (stride 2). Per-vertex color and
+	// UVs were previously interleaved here at VERTEX_SIZE = 2 + 2 + 4, but
+	// canvas2d has no per-pixel vertex-attribute interpolation: the four
+	// color floats per vertex were computed and never read by drawTriangle,
+	// and UVs come straight from `sequence.getUVs()` (a spine-cached array
+	// keyed by sequence frame, stable per-frame). Dropping both halves the
+	// vertex-buffer traffic and cuts the helper code to a single
+	// `computeWorldVertices(..., stride: 2)` call per attachment.
+	static VERTEX_SIZE = 2;
 
 	private ctx: CanvasRenderingContext2D;
 
 	public triangleRendering = false;
 	public debugRendering = false;
-	private vertices = Utils.newFloatArray(8 * 1024);
+	private vertices = Utils.newFloatArray(2 * 1024);
 	private tempColor = new Color();
+
+	// Clipping state. `clipper` tracks the active clip range (which `endSlot`
+	// it ends at) per spine-core's polygon-clip helper; `clippingVerts` is
+	// the world-space polygon buffer; `clipped` tells us whether we've
+	// `ctx.save()`d to install a canvas clip so we know to `restore()` when
+	// the range closes.
+	private clipper = new SkeletonClipping();
+	private clippingVerts: NumberArrayLike = Utils.newFloatArray(8);
+	private clipped = false;
 
 	constructor (context: CanvasRenderingContext2D) {
 		this.ctx = context;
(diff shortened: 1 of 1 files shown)


davidetan:
Hello and thanks for submitting this, and for your contribution :)

We have added this review to our `spine-ts` queue. There are a couple of thing first, so it'll take a while before we can review it.

obiot:
> Hello and thanks for submitting this, and for your contribution :)
> 
> We have added this review to our `spine-ts` queue. There are a couple of thing first, so it'll take a while before we can review it.

Thanks, and no worries, in the mean time we still have our "custom" canvas skeleton renderer working perfectly in our own plugin !

davidetan:
@obiot Thanks for the PR.

I did not merge the patch directly and instead created the https://github.com/EsotericSoftware/spine-runtimes/tree/canvas-clip-blend-modes branch.
I did that because the clipping implementation used Canvas ctx.clip() as a clip range, which only approximates Spine clipping and has edge cases with inverse clipping, inactive slots, nested/restarted clipping, and Canvas state restoration.

I reimplemented the feature around `SkeletonClipping` and `SkeletonRendererCore`, matching the WebGL clipping lifecycle, and simplyfing the triangle render code.

The final result keeps the intent of the PR but avoids canvas-specific clipping state and shares the more correct runtime clipping implementation.

Let me know if the implementation fits your need, or if we need to make some improvements.

obiot:
Hi @davidetan  thank you for reviewing it, and yes your version is definitely a better version on the clipping side !

As for using it for melonJS when falling back to canvas, not yet. To fully move away from our "custom" CanvasSkeletonRenderer we would also need a way to have some passthrough so that the Spine plugin can draw through melonJS (else we lose tinting, masking and camera transforms on a Spine Object). it works in WebGL because Spine exposes the WebGL geometry that we are pushing directly to our Batcher.  But maybe for the next PR I'll try to propose something :)

obiot:
Hi @davidetan, any update on this one ? thank you

davidetan:
@obiot Hey, sorry for the delay! 

However, I'm worried that what you're asking for would make `spine-canvas` more complex than it should be.

If you just need geometry data, you can directly use the renderer-agnostic `SpineRendererCore`, exactly as I did in the branch mentioned above (and in many other runtimes): https://github.com/EsotericSoftware/spine-runtimes/blob/canvas-clip-blend-modes/spine-ts/spine-core/src/SkeletonRendererCore.ts

We haven't documented it yet, but it's super easy to understand. You don't even have to worry about slots, clipping, or all the other internals. It directly provides you with a list of batches in the form of commands containing the respective geometry data, blend modes, textures, ...

Look at the `drawTriangles` method and how trivial it is:
https://github.com/EsotericSoftware/spine-runtimes/blob/canvas-clip-blend-modes/spine-ts/spine-canvas/src/SkeletonRenderer.ts#L132

What do you think about this approach?

obiot:
hi @davidetan I definitely don't disagree, my only goal was to bring feature parity to the canvas renderer as some of the examples were failing to render properly, and all this without having to implement a full custom skeletonRenderer on our side (as you know the less custom code we have, the less maintenance effort we need on it) :
https://github.com/melonjs/melonJS/blob/master/packages/spine-plugin/src/SkeletonRenderer.js

But I'll definitely look at the `SkeletonRendererCore` API also because we recently added WebGPU rendering to our Spine Plugin, and it also required some more custom code.... 

nevertheless, if the same result can be achieved with the new API, im definitely all for it !

davidetan:
> hi @davidetan I definitely don't disagree, my only goal was to bring feature parity to the canvas renderer as some of the examples were failing to render properly, and all this without having to implement a full custom skeletonRenderer on our side (as you know the less custom code we have, the less maintenance effort we need on it) : https://github.com/melonjs/melonJS/blob/master/packages/spine-plugin/src/SkeletonRenderer.js
> 
> But I'll definitely look at the `SkeletonRendererCore` API also because we recently added WebGPU rendering to our Spine Plugin, and it also required some more custom code....
> 
> nevertheless, if the same result can be achieved with the new API, im definitely all for it !

Then I think that the `SkeletonRendererCore` API is the best choice. You can also have a look at `spine-canvaskit` and `spine-construct3` to understand how it's used.

At this point, I think we can close this PR, and I can instead merge the `spine-canvas` branch that uses `SkeletonRendererCore`.

If you have any questions regarding the `SkeletonRendererCore` API, you can open a forum thread, and we will be more than happy to provide the support you need to use it in MelonJS :)

PR #3129, opened 2026-06-17
fix for godot extension not building correctly

wyldspace:
Fixed incorrect location for gdextension file when bundling and uploading to s3


badlogic:
The extension builds correctly.

PR #3133, opened 2026-06-23
Bugfixes

wyldspace:

[Godot] Bug fixes, performance improvements, and documentation updates
Bug fixes
SpineSkin.get_attachment() always returned null — inverted null check caused the function to return null when an attachment was found and crash when it wasn't.

SpineSkeleton.get_ik_constraints(), get_transform_constraints(), get_path_constraints() returned wrong types — a double-negated RTTI filter (!X == false) caused the functions to skip matching constraints and process wrong-type ones instead, producing invalid wrapper objects and crashes on property access. Array packing was also broken (result[i] instead of result[size]), leaving null entries when matching constraints weren't at index 0.

SpineSprite.draw(), get_global_bone_transform(), set_global_bone_transform() could dereference an invalid skeleton — validity guards used && instead of ||, only bailing out when both refs were invalid. Since animation_state is unused in these functions, the guards now check only skeleton.

SpineSlotNode signal not disconnected on unparent (Godot 4) — NOTIFICATION_UNPARENTED fires after the parent pointer is cleared in Godot 4, so get_parent() returned null and the world_transforms_changed disconnect was never called, leaving a dangling connection. The parent is now stored at connect time and used at disconnect time.

SpineObjectWrapper signal connection outlived the wrapper (Godot 3) — in Godot 3, signal connections hold a raw Object* with no reference bump. Dropping the last Ref<> to a wrapper while still connected freed the wrapper but left a live connection on the owner, causing a crash on the next signal emission. A destructor now safely disconnects using ObjectDB::get_instance to guard against an already-freed owner.

Stale bone/slot cache after skeleton rebuild — SpineSkeleton._cached_bones and _cached_slots were not cleared when set_spine_sprite() deleted and rebuilt the skeleton. A new skeleton could reuse the same memory addresses, causing find_bone() and find_slot() to return stale wrappers pointing to freed spine-cpp objects.

Performance improvements
sort_slot_nodes() — reduced per-frame scene-tree churn — the function now early-exits when no user children exist beyond the mesh instances. move_child() is skipped when a node is already at the correct index. set_draw_behind_parent() is skipped when already set. Together these eliminate redundant scene-tree notifications on every frame when draw order is stable.

Debug overlay font lookup — eliminated per-frame heap allocation — the editor hover tooltip previously allocated and immediately freed a Control node on every draw call to obtain the default font. The font is now fetched once and cached for the lifetime of the sprite.

Documentation
Updated 12 XML API documentation files to match the current C++ implementation:

SpineSlot — complete rewrite; the documented flat API (color, attachment, deform getters/setters) was replaced in a prior refactor by get_pose() and get_applied_pose() returning a SpineSlotPose, but the doc was never updated.
SpinePathConstraint — complete rewrite; mix/position/spacing methods removed, pose-based API documented.
SpineTrackEntry — get_attachment_threshold and get_draw_order_threshold renamed to their correct get_mix_* names; get_hold_previous (not bound) removed; 7 missing methods added including get_additive, get_mix_interpolation, was_applied, and set_mix_duration_and_delay.
SpineTransformConstraint — get_target/set_target corrected to get_source/set_source; 6 unbound mix setter/getter pairs removed; get_pose/get_applied_pose and the skeleton argument on update() added.
SpineSkeleton — physics argument added to update_world_transform; 9 missing methods added (find_physics_constraint, find_slider, get_physics_constraints, get_sliders, get_time, set_time, update, physics_translate, physics_rotate); "worldX" typo in get_y/set_y descriptions corrected.
SpineConstant — MixInterpolation (5 values) and ScaleYMode (3 values) enums were missing entirely.
SpineAnimationState — get_current corrected to get_track; clear_track argument name fixed from arg0 to track_id.
SpineSkeletonDataResource — 5 missing methods added: find_physics_constraint_data, get_physics_constraints, get_reference_scale, set_reference_scale, update_skeleton_data.
SpineAnimation — apply() signature corrected from 8 arguments to 10.
SpineSprite — time_scale, debug_root, and debug_root_color properties were missing from the members section.
SpineBone — set_transform and set_global_transform argument names corrected to match their bindings; unbound apply_world_transform_2d removed.
SpineSliderData — 4 missing methods added: get_max, set_max, get_local, set_local.

### DIFF
--- a/.github/actions/setup-godot-deps-3/action.yml
+++ b/.github/actions/setup-godot-deps-3/action.yml
@@ -3,7 +3,7 @@ description: Setup python, install the pip version of scons.
 inputs:
   python-version:
     description: The python version to use.
-    default: "3.x"
+    default: "3.12"
   python-arch:
     description: The python architecture.
     default: "x64"
--- a/.github/actions/setup-godot-deps-3/action.yml
+++ b/.github/actions/setup-godot-deps-3/action.yml
@@ -3,7 +3,7 @@ description: Setup python, install the pip version of scons.
 inputs:
   python-version:
     description: The python version to use.
-    default: "3.12"
+    default: "3.x"
   python-arch:
     description: The python architecture.
     default: "x64"
@@ -23,7 +23,7 @@ runs:
       shell: bash
       run: |
         python -c "import sys; print(sys.version)"
-        python -m pip install scons==4.4.0
+        python -m pip install scons==4.7.0
         scons --version
 
     - name: Setup dotnet 6
--- a/.github/actions/setup-godot-deps-4/action.yml
+++ b/.github/actions/setup-godot-deps-4/action.yml
@@ -3,7 +3,7 @@ description: Setup python, install the pip version of scons.
 inputs:
   python-version:
     description: The python version to use.
-    default: "3.x"
+    default: "3.12"
   python-arch:
     description: The python architecture.
     default: "x64"
--- a/.github/actions/setup-godot-deps-4/action.yml
+++ b/.github/actions/setup-godot-deps-4/action.yml
@@ -3,7 +3,7 @@ description: Setup python, install the pip version of scons.
 inputs:
   python-version:
     description: The python version to use.
-    default: "3.12"
+    default: "3.x"
   python-arch:
     description: The python architecture.
     default: "x64"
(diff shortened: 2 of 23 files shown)


badlogic:
Cheers, I picked out most of the fixes.

wyldspace:
glad to be of help.

PR #3137, opened 2026-06-29
[ios] feat(spine-ios): add per-view texture filter (nearest / linear)

Dwarven:
Add fragmentShaderLinear and linearTextureFilter on SpineUIView and SpineView. SpineRenderer selects the fragment function at pipeline creation time. Default remains nearest for backward compatibility.

### Summary

This PR adds **per-view texture filter configuration** to spine-ios Metal rendering, allowing callers to choose between **nearest** (default) and **linear** without changing global behavior.

### Motivation

Different use cases may require different texture filters: some content looks best with **nearest**, others with **linear**. The current Metal shader hard-codes `nearest`, and there is no public API to override it per `SpineUIView`.

### Solution

1. **Metal shaders** — Keep `fragmentShader` (nearest) and add `fragmentShaderLinear` (linear).
2. **`SpineRenderer`** — Accept `textureFilter: TextureFilter` (default `.nearest`) and select the fragment function at pipeline creation time.
3. **`SpineUIView`** — Expose `linearTextureFilter: Bool` (default `false`) on initializers; map to `TextureFilter` internally. Uses `Bool` for `@objc` interoperability.
4. **`SpineView` (SwiftUI)** — Forward `linearTextureFilter` to `SpineUIView`.

### API

```swift
let view = SpineUIView(from: source)                          // nearest (default)
let view = SpineUIView(from: source, linearTextureFilter: true) // linear
SpineView(from: source, linearTextureFilter: true)
```

Objective-C: `linearTextureFilter:YES` on `@objc` initializers.

### Changed files

| File | Change |
|------|--------|
| `spine-ios/Sources/SpineiOS/Metal/SpineShaders.metal` | Add `fragmentShaderLinear` |
| `spine-ios/Sources/SpineiOS/Metal/SpineRenderer.swift` | `textureFilter` init parameter; dynamic fragment function name |
| `spine-ios/Sources/SpineiOS/SpineUIView.swift` | `linearTextureFilter: Bool` property and initializer parameters |
| `spine-ios/Sources/SpineiOS/SpineView.swift` | Pass `linearTextureFilter` through to `SpineUIView` |

### Backward compatibility

- Default remains **`nearest`** (`linearTextureFilter: false`).
- No changes required for existing call sites.

- [x] I confirm this contribution is made under the Esoteric Software LLC [CLA](http://esotericsoftware.com/licenses/cla.txt).

### DIFF
--- a/spine-ios/Sources/SpineiOS/Metal/SpineRenderer.swift
+++ b/spine-ios/Sources/SpineiOS/Metal/SpineRenderer.swift
@@ -81,7 +81,8 @@ internal final class SpineRenderer: NSObject, MTKViewDelegate {
         commandQueue: MTLCommandQueue,
         pixelFormat: MTLPixelFormat,
         atlasPages: [UIImage],
-        pma: Bool
+        pma: Bool,
+        textureFilter: TextureFilter = .nearest
     ) throws {
         self.device = device
         self.commandQueue = commandQueue
@@ -95,6 +96,7 @@ internal final class SpineRenderer: NSObject, MTKViewDelegate {
         #endif
 
         let defaultLibrary = try device.makeDefaultLibrary(bundle: bundle)
+        let fragmentFunctionName = Self.fragmentFunctionName(for: textureFilter)
         let textureLoader = MTKTextureLoader(device: device)
         textures =
             try atlasPages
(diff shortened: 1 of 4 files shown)


badlogic:
Thanks! I ended up implementing this in a slightly different way.

Rather than selecting between two fragment shaders and continuing to hardcode nearest as the default, the Metal renderer now creates a sampler state for each atlas page. By default it honors the page's exported minification, magnification, mipmap, and wrap settings.

I also added an Objective-C-compatible `SpineTextureFilter` API to `SpineUIView` and `SpineView`:

- `.atlas` uses the atlas page settings and is the default
- `.nearest` overrides filtering for every page in the view
- `.linear` overrides filtering for every page in the view

The existing Objective-C initializer selectors are preserved, with additional overloads for the filter override. There is also a new interactive `TextureFiltering.swift` example, plus README, changelog, and website documentation updates.

This fixes the underlying issue—the runtime was ignoring the filters already exported in the atlas—while still supporting the per-view nearest/linear control proposed here. Thanks for identifying the hardcoded sampler and putting together the initial implementation!

Issue #3155, opened 2026-07-24
[ios] Swift Package Compilation Error


Dwarven:
This diff in `Package.swift` caused a compilation error.

https://github.com/EsotericSoftware/spine-runtimes/commit/4f0728e2a1cfee8320a2ac38930ccf22f21d0fd4#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR47

```diff
+            resources: [
+                .process("Metal/SpineShaders.metal")
+            ]
```

```sh
error: Unexpected duplicate tasks
    note: Target 'spine-ios_SpineiOS' (project 'spine-ios') has compile command with input '/path/to/project/Tuist/.build/checkouts/spine-runtimes/spine-ios/Sources/SpineiOS/Metal/SpineShaders.metal'
    note: Target 'spine-ios_SpineiOS' (project 'spine-ios') has compile command with input '/path/to/project/Tuist/.build/checkouts/spine-runtimes/spine-ios/Sources/SpineiOS/Metal/SpineShaders.metal'
error: Unexpected duplicate tasks
    note: Target 'spine-ios_SpineiOS' (project 'spine-ios'): MetalLink /Users/xxx/Library/Developer/Xcode/DerivedData/project-doyabcwxkinldxgvmmaeyutbdmzi/Build/Products/Debug-iphoneos/spine-ios_SpineiOS.bundle/default.metallib
    note: Target 'spine-ios_SpineiOS' (project 'spine-ios'): MetalLink /Users/xxx/Library/Developer/Xcode/DerivedData/project-doyabcwxkinldxgvmmaeyutbdmzi/Build/Products/Debug-iphoneos/spine-ios_SpineiOS.bundle/default.metallib
```

PR #3156, opened 2026-07-28
[ios] Fix duplicate Metal shader compilation in Package.swift

Dwarven:
This PR fixes a build error #3155  where `SpineShaders.metal` was being compiled twice - once as source code and once as a resource file, causing "Unexpected duplicate tasks" errors in Xcode.

### Problem

When the Metal shader was declared in the `resources` section, SPM/Xcode tried to process it as a resource file. However, .metal files in the source path are automatically compiled as source code by the Metal compiler, resulting in duplicate compilation tasks:

```sh
error: Unexpected duplicate tasks
  note: Target 'spine-ios_SpineiOS' has compile command with input 'SpineShaders.metal'
  note: Target 'spine-ios_SpineiOS' has compile command with input 'SpineShaders.metal'
```

### Solution

Removed the `resources` section from the `SpineiOS` target definition. Metal shader files (`.metal`) are now correctly handled as source files only, not as resources.

### Changes

- Removed `resources: [.process("Metal/SpineShaders.metal")]` from `SpineiOS` target

- The Metal shader is now automatically compiled as source code (the correct behavior)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -42,10 +42,7 @@ let package = Package(
                 "SpineSwift",
                 "SpineShadersStructs",
             ],
-            path: "spine-ios/Sources/SpineiOS",
-            resources: [
-                .process("Metal/SpineShaders.metal")
-            ]
+            path: "spine-ios/Sources/SpineiOS"
         ),
         .target(
             name: "SpineC",


Dwarven:
@badlogic `.metal` files are automatically compiled as source code by the Metal compiler, **NOT** resources.

PR #3164, opened 2026-08-30
[haxe] Fixed timeline property id corruption caused by Haxe 4 flash target rest-spread bug

mallyskies:
Timeline's constructors forwarded a rest argument with super(frameCount, ...propertyIds). The Haxe compiler miscompiles a call-site spread on the flash (AVM2) target: genswf9 built the expanded argument list as [rest...] @ List.rev fixed_args, so the fixed arguments arrived reversed with the rest items prepended. frameCount therefore arrives as a property id and propertyIds as [frameCount], with no compile error and no runtime exception -- property ids collide across timelines and AnimationState mixing corrupts. Fixed upstream in HaxeFoundation/haxe#13017, which is merged to development and so lands in Haxe 5 once it is GA.

Revised after @davidetan's review. The first version declared propertyIds as Array<String>, which changed the public constructor signatures and would have broken any user-defined timeline subclass. This version takes the approach he suggested -- call super() without the ids and assign propertyIds afterwards, which Timeline itself already does -- and changes no signatures anywhere.

Two things make it smaller than the original description implied. Only three call sites in the runtime actually spread: CurveTimeline, SlotCurveTimeline and DrawOrderFolderTimeline. The rest pass a fixed argument list into a rest parameter, which the compiler handles correctly and which was never affected -- so they need no change. And the workaround is guarded by #if (flash && !haxe5), with the original line unchanged in the #else, so every other target and Haxe 5 onward compile exactly what they do today. Three files, +15 lines, nothing deleted.

Verification:

compiles for -swf, -js, and -swf -D haxe5 (which type-checks the #else branch on the flash target)
propertyIds and frames identical to current 4.3 across 12 timeline types, with the guarded branch forced on a runnable target
haxelib run formatter --check clean
decompiled from a built SWC and confirmed in a shipping AIR app on Android: super(param1); propertyIds = rest.copy();, no spread and no length-dispatch chain. Reel symbols and animation mixing, both broken before, are correct.
While the conditional isn't strictly necessary, it serves as a monument to clarify the scope of the defect and avoid someone unknowingly reintroducing the defect.

### DIFF
--- a/spine-haxe/spine-haxe/spine/animation/CurveTimeline.hx
+++ b/spine-haxe/spine-haxe/spine/animation/CurveTimeline.hx
@@ -43,7 +43,12 @@ abstract class CurveTimeline extends Timeline {
 	/** @param bezierCount The maximum number of Bezier curves. See CurveTimeline.shrink().
 	 * @param propertyIds Unique identifiers for the properties the timeline modifies. */
 	public function new(frameCount:Int, bezierCount:Int, propertyIds:...String) {
+		#if (flash && !haxe5) // See HaxeFoundation/haxe#13017
+		super(frameCount);
+		this.propertyIds = propertyIds;
+		#else
 		super(frameCount, ...propertyIds);
+		#end
 		curves = ArrayUtils.resize(new Array<Float>(), frameCount + bezierCount * BEZIER_SIZE, 0);
 		curves[frameCount - 1] = STEPPED;
 	}
--- a/spine-haxe/spine-haxe/spine/animation/DrawOrderFolderTimeline.hx
+++ b/spine-haxe/spine-haxe/spine/animation/DrawOrderFolderTimeline.hx
@@ -44,7 +44,12 @@ class DrawOrderFolderTimeline extends Timeline {
 	/** @param slots spine.Skeleton.slots indices controlled by this timeline, in setup order.
 	 * @param slotCount The maximum number of slots in the skeleton. */
 	public function new(frameCount:Int, slots:Array<Int>, slotCount:Int) {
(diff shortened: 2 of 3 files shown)


davidetan:
This changes the public API, and we can't do that. I think the only viable solution while waiting for the new Haxe release is to assign the `propertyIds` in the constructor rather than relying on the `super` call.

For example, changing:

```haxe
public function new(frameCount:Int, bezierCount:Int, slotIndex:Int, propertyIds:...String) {
       super(frameCount, bezierCount, ...propertyIds);
       this.slotIndex = slotIndex;
}
```

to:

```haxe
public function new(frameCount:Int, bezierCount:Int, slotIndex:Int, propertyIds:...String) {
       super(frameCount, bezierCount);
       this.propertyIds = propertyIds;
       this.slotIndex = slotIndex;
}
```

That said, I’d prefer to avoid changing the runtime to work around a Haxe issue that has already been addressed upstream and is expected to be included in an upcoming release.

mallyskies:
Thanks for your suggestions! I reworked along the lines you suggested and force-pushed; the diff is now three files, +15 lines, no signature changes.

Only three call sites in the runtime actually need the flash target spread workaround: CurveTimeline, SlotCurveTimeline and DrawOrderFolderTimeline. Everything else passes a fixed argument list into a rest parameter, which the compiler handles correctly and was never affected.

While I understand the reluctance to carry a workaround for a compiler bug that is already fixed upstream, keep in mind that Haxe 5 is prerelease and has been for over a year. And even once it's released, it may take time for major libraries like Flixel, OpenFL, and Heaps to catch up. I expect most production Haxe software will be built on Haxe 4.3 for quite some time.

PR #3166, opened 2026-09-01
[godot] Resolves #2735, resolves Godot Editor crash when a preview animation is set to an invalid animation

### LINKED ISSUE #2735: [godot] Editor crash when loading a scene with non-existent animation set as preview-animation
Hey!

When in you open a godot tscn scene file where you have a `SpineSprite` which has the `preview_animation` property set to an animation name that doesn't exist, the editor will crash.

Easiest way to repro this is to just go into the TSCN scene file and change the `preview_animation` property to some garbage, and then open that scene from the godot editor. That'll cause a crash and a core dump on both linux and windows. Tested on `Godot Engine v4.3.stable.custom_build (77dcf97d82cbfe4e4615475fa52ca03da645dbd8)` (custom spine runtime godot build).

I think it should just rather set the `preview_animation` to `-- Empty --` (but not write the scene file) and probably just write an error/warning to the console/debugger.

lukeeingram:


lukeeingram:
Related to #2735

badlogic:
cheers!

